PR #44, opened 2019-05-27
ATS-434 : Implement `/info` endpoint in T-Engines

DenisGabriela:
- implement '/info' endpoint
   - add engine_config files
   - use SNAPSHOT transform-model with new Transform Config models (TODO update after transform-model release)
   - remove 'tests' from travis stages
   - add new junits
   - add test resources


alfresco-transformer-base/src/test/java/org/alfresco/transformer/AbstractTransformerControllerTest.java
```diff
@@ -38,18 +37,24 @@
 import java.net.URL;
 import java.nio.channels.FileChannel;
 import java.nio.file.Files;
+import java.util.*;
```
tzclucian: We should not use `*` imports.

montgolfiere:
I notice we have default maxSourceSizeBytes for most of the Libreoffice transforms & also for one of the Tika transforms.

For consistency, could (should ?) we also take the opportunity to put a tested default maxSourceSizeBytes (even if slightly arbitrary) on all of the other transforms ? Over time, based on feedback (and further testing in specific environments) we can tweak up or down, as & when needed.

alfresco-transformer-base/src/test/java/org/alfresco/transformer/AbstractTransformerControllerTest.java
```diff
@@ -287,4 +299,134 @@ public void testEmptyPojoTransform() throws Exception
...
+                header().string(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_UTF8_VALUE))
+            .andReturn().getResponse().getContentAsString();
+
+        TransformConfig transformConfig = objectMapper.readValue(response, TransformConfig.class);
+
+        assertEquals(expectedTransformConfig, transformConfig);
+    }
+
+    @Test
+    public void testGetInfoFromConfigWithDuplicates() throws Exception
```
montgolfiere: Is this a negative test ? If so shouldn't it result in an error ? 

Or otherwise, will the T-Engine show warnings in the log (that indicate duplicates in the config) ?

alfresco-transformer-base/src/test/java/org/alfresco/transformer/AbstractTransformerControllerTest.java
```diff
@@ -287,4 +299,134 @@ public void testEmptyPojoTransform() throws Exception
...
+        assertEquals(3, transformConfig.getTransformOptions().get("engineXOptions").size());
+        assertEquals(1,
+            transformConfig.getTransformers().get(0).getSupportedSourceAndTargetList().size());
+        assertEquals(1,
+            transformConfig.getTransformers().get(0).getTransformOptions().size());
+
+    }
+
+    @Test
+    public void testGetInfoFromIncompleteConfig() throws Exception
```
montgolfiere: Add comment to explain what "incomplete" means ?

Again, if a negative test shouldn't it result in a (startup) error ?

alfresco-docker-libreoffice/src/main/resources/engine_config.json
```diff
@@ -0,0 +1,18 @@
+{
+  "transformers": [
+    {
+      "transformerName": "libreoffice",
+      "supportedSourceAndTargetList": [
+        {"sourceMediaType": "application/msword",                                                                                        "targetMediaType": "application/msword" },
```
montgolfiere: I notice we have default maxSourceSizeBytes for most of the LibreOffice transforms & also for one of the Tika transforms.

For consistency, could (should ?) we also take the opportunity to put a tested default maxSourceSizeBytes (even if slightly arbitrary) on all of the other transforms ? Over time, based on feedback (and further testing in specific environments) we can tweak up or down, as & when needed.

CezarLeahu: We've added the `maxSourceSizeBytes` based on the values in the *alfresco-repository* JSON config: [transform-service-config.json](https://github.com/Alfresco/alfresco-repository/blob/master/src/main/resources/alfresco/transform-service-config.json).

We don't have any information about other transformations. We should raise a separate story for figuring out what the proper limits would be - may the Repo team can help with that.

montgolfiere: OK - let's make this a separate task.

alfresco-transformer-base/src/test/java/org/alfresco/transformer/AbstractTransformerControllerTest.java
CezarLeahu: No, this is the way the serialization libraries (_Jackson_, in our case) usually work: they deduplicate entries when deserializing a **JSON array** into a **Java Set<>**. They don't throw any errors.

montgolfiere: Hmm - we may need a way (in future ?) to provide some level of *semantic* validation to help custom configurations.

alfresco-transformer-base/src/test/java/org/alfresco/transformer/AbstractTransformerControllerTest.java
CezarLeahu: Some TEngines can have no `transformOptions` (e.g. _LibreOffice_); I think this is the case with this particular test. Maybe we should rename the test.

alfresco-transformer-base/src/test/java/org/alfresco/transformer/AbstractTransformerControllerTest.java
CezarLeahu: It might not be worth it. Unless if you see this scenario happening often and causing problems, I don't see it as a priority.
We're definitely not implementing it in this story, as that validation would take quite a bit of work (it doesn't match our Java model, we'd have to parse the `JSONObject`s).

montgolfiere: Ok, let's see if it becomes an issue. In the meantime, we should document the current behaviour (as demonstrated by the tests).

CezarLeahu:
We've addressed most of the CR comments.

I'll be merging this PR once the build is green.

PR #45, opened 2019-05-28
REPO-4331: Add remaining core transformers

eknizat:
* HtmlParserContentTransformer
* AppleIWorksContentTransformer
* StringExtractingContentTransformer
* TextToPdfContentTransformer
* OOXMLThumbnailContentTransformer (disabled as it doesn't work with Java 11)


montgolfiere:
I haven't had a chance to review the code. 

In the meantime, a minor "naming" comment ... I wonder if we should shorten and rename both the folder and docker image name to prefix with "-"transform". Hence, maybe something like:

Folder: alfresco-docker-transform-misc
Image: alfresco-transform-misc

This would:
- be more consistent with the associated repo / project names
- also give us the option to potentially rename the other T-engines in the future

alfresco-docker-miscellaneous-transformers/src/main/java/org/alfresco/transformer/MiscellaneousTransformersController.java
```diff
@@ -0,0 +1,152 @@
...
+{
+    private static final Logger logger = LoggerFactory.getLogger(MiscellaneousTransformersController.class);
+
+    @Autowired
+    private SelectingTransformer transformer;
+
+    public MiscellaneousTransformersController()
+    {
+        logger.info("--------------------------------------------------------------------------------------------------------------------------------------------------------------");
+        Arrays.stream(ENTERPRISE_LICENCE.split("\\n")).forEach(logger::info);
```
alandavis: Please double check this code has not changed, as the T-Engines have recently been opened up to be not enterprise only.

alfresco-docker-miscellaneous-transformers/src/main/java/org/alfresco/transformer/transformers/AppleIWorksContentTransformer.java
```diff
@@ -0,0 +1,176 @@
...
+
+    @Autowired
+    public AppleIWorksContentTransformer(SelectingTransformer miscTransformer)
+    {
+        super(miscTransformer);
+    }
+
+    @Override
+    public boolean isTransformable(String sourceMimetype, String targetMimetype, Map<String, String> parameters)
+    {
```
alandavis: If I remember correctly the new transformers don't do isTransformable in Java code. It should be checked via configuration. This appears to be code copied from the legacy transformers. As a result the static finals above can also be removed.

.travis.yml
```diff
@@ -25,6 +25,9 @@ branches:
     - release/SP/2.0.N
     - company_release
     - /^ATS-.*$/
+    - release_repo_4331
```
alandavis: This will need to be removed before the merge.

alfresco-docker-transform-misc/Dockerfile
```diff
@@ -0,0 +1,35 @@
+# Image provides a container in which to run miscellaneous transformations for Alfresco Enterprise Content Services.
+
+# The container is only intended to be used with the Alfresco Enterprise editon which is covered by https://www.alfresco.com/legal/agreements and https://www.alfresco.com/terms-use.
```
alandavis: This line was added for Tika. It should not be here but there may be other licenses for 3rd part jars the the 'misc' T-Engine uses.

alfresco-docker-transform-misc/LICENSES.md
```diff
@@ -0,0 +1,11 @@
+### Licenses
+
```
alandavis: If you copied this from the other T-Engines, please raise an ATS jira to say that this should be changed in all 5 standard T-Engines as they will be used in community now. If this has changed in the other T-Engines, please make the text the same. I know the InT team were changing licenses a while back.

alfresco-docker-transform-misc/src/main/java/org/alfresco/transformer/MiscController.java
```diff
@@ -0,0 +1,152 @@
+/*
+ * #%L
+ * Alfresco Transform Core
+ * %%
+ * Copyright (C) 2005 - 2019 Alfresco Software Limited
+ * %%
+ * This file is part of the Alfresco software.
+ * -
```
alandavis: Again these license headers look wrong for community code. Add to the ATS issue or change to be the same as the other T-Engines.

alfresco-docker-transform-misc/src/main/java/org/alfresco/transformer/MiscController.java
```diff
@@ -0,0 +1,152 @@
...
+{
+    private static final Logger logger = LoggerFactory.getLogger(MiscController.class);
+
+    @Autowired
+    private SelectingTransformer transformer;
+
+    public MiscController()
+    {
+        logger.info("--------------------------------------------------------------------------------------------------------------------------------------------------------------");
+        Arrays.stream(ENTERPRISE_LICENCE.split("\\n")).forEach(logger::info);
```
alandavis: No longer an Enterprise license.

alfresco-docker-transform-misc/src/main/java/org/alfresco/transformer/transformers/AppleIWorksContentTransformer.java
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * <p>
+ * Since HTML Parser was updated from v1.6 to v2.1, META tags
+ * defining an encoding for the content via http-equiv=Content-Type
+ * will ONLY be respected if the encoding of the content item
+ * itself is set to ISO-8859-1.
+ * </p>
+ *
+ * <p>
+ * Tika Note - could be converted to use the Tika HTML parser,
```
alandavis: This class header appears to have been copied from the Tika transformer. Other transformers have the same problem (not adding this comment to all of them). I suggest you use class header text from the legacy transformers, having checked it still makes sense.

alfresco-docker-transform-misc/src/main/java/org/alfresco/transformer/transformers/OOXMLThumbnailContentTransformer.java
```diff
@@ -0,0 +1,159 @@
...
+                throw new Exception("No thumbnail present in file, unable to generate " + targetMimetype);
+            }
+        }
+        catch (IOException e)
+        {
+            throw new AlfrescoRuntimeException("Unable to transform file.", e);
+        }
+    }
+
+    /*
```
alandavis: What needs to be done here? If it cannot be done now is there an ATS issue?

alfresco-docker-transform-misc/src/main/java/org/alfresco/transformer/transformers/AppleIWorksContentTransformer.java
eknizat: I'll updated the java doc fo this method, good catch. But the headers are already the same as the ones in alfresco-repository

CezarLeahu:
I've updated the copyright headers on all the *.java files in the project. We'll fix the other license issues in the story raised by Erik (ATS-460 - Dockerfiles, LICENSES.md files, controller log messages).

It would be nice to address the _OOXML_ issues before the merge, but nonetheless, they can be fixed in a subsequent PR.

Everything else looks good to me. I've approved the PR.

eknizat:
Thanks Cezar.
I have created a repo ticket to fix the OOXML issue: REPO-4473.

alfresco-docker-miscellaneous-transformers/src/main/java/org/alfresco/transformer/MiscellaneousTransformersController.java
montgolfiere: Erik raised here: https://issues.alfresco.com/jira/browse/ATS-460

PR #46, opened 2019-06-04
ATS-434 : Implement `/transform/config` endpoint in T-Engines

DenisGabriela:
- rename endpoint since '/info' is already used by Spring Boot
   - fix config file loading


CezarLeahu:
I've updated the design docs: [Alfresco/alfresco-transform-service/pull/124](https://github.com/Alfresco/alfresco-transform-service/pull/124).

PR #51, opened 2019-06-06
ATS-250 : ATS: T-Engines: Improve the convert message error handling (T-Base)

DenisGabriela:
- code refactoring:
 -- if successful -> return the object ( Optional.ofNullable(request) )
 -- if not successful -> throw an exception (instead of sending an error message on the queue)
   - when calling 'convert()' method, catch this exception and send a T-Reply with error message on the queue

### DIFF
--- a/alfresco-transformer-base/src/main/java/org/alfresco/transformer/QueueTransformService.java
+++ b/alfresco-transformer-base/src/main/java/org/alfresco/transformer/QueueTransformService.java
@@ -104,9 +104,9 @@ public class QueueTransformService
         }
         catch (TransformException e)
         {
+            logger.error(e.getMessage(), e);
             replyWithError(replyToDestinationQueue, HttpStatus.valueOf(e.getStatusCode()),
                 e.getMessage(), correlationId);
-            logger.error(e.getMessage(), e);
             return;
         }
 
--- a/alfresco-transformer-base/src/main/java/org/alfresco/transformer/QueueTransformService.java
+++ b/alfresco-transformer-base/src/main/java/org/alfresco/transformer/QueueTransformService.java
@@ -95,12 +95,12 @@ public class QueueTransformService
             return;
         }
 
-        logger.info("New T-Request from queue with correlationId: {0}", correlationId);
+        logger.info("New T-Request from queue with correlationId: {}", correlationId);
 
         Optional<TransformRequest> transformRequest;
         try
         {
-            transformRequest = convert(msg);
+            transformRequest = convert(msg, correlationId);
         }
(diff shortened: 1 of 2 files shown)


alfresco-transformer-base/src/main/java/org/alfresco/transformer/QueueTransformService.java
```diff
@@ -94,75 +97,65 @@ public void receive(final Message msg)
...
+            return;
+        }
 
-        TransformRequest transformRequest = convert(msg, correlationId, replyToDestinationQueue);
-        if (transformRequest == null)
+        if (!transformRequest.isPresent())
         {
-            logger.error("Exception during T-Request deserialization! T-Reply with error has been "
-                + "sent to T-Router!");
+            logger.error("T-Request is null deserialization!");
```
montgolfiere: Error messages should contain some detailed context (correlationId, requestId, sourceReference, ...) to allow log grep / tracing.

alfresco-transformer-base/src/main/java/org/alfresco/transformer/QueueTransformService.java
```diff
@@ -94,75 +97,65 @@ public void receive(final Message msg)
...
-                logger.error("T-Request is null deserialization!");
-                replyWithInternalSvErr(destination,
-                    "JMS exception during T-Request deserialization: ", correlationId);
-            }
-            return request;
+            return Optional.ofNullable(request);
         }
         catch (MessageConversionException e)
         {
             String message = "Message conversion exception during T-Request deserialization: ";
```
montgolfiere: See comment above - add T-Request context to error messages ?

alfresco-transformer-base/src/main/java/org/alfresco/transformer/QueueTransformService.java
```diff
@@ -94,75 +97,65 @@ public void receive(final Message msg)
...
-            logger.error(message + e.getMessage(), e);
-            replyWithBadRequest(destination, message + e.getMessage(), correlationId);
+            throw new TransformException(HttpStatus.BAD_REQUEST.value(), message + e.getMessage());
         }
         catch (JMSException e)
         {
             String message = "JMS exception during T-Request deserialization: ";
```
montgolfiere: See comment above (add T-Request context to error messages).

DenisGabriela:
@montgolfiere , thanks for review. I've updated the PR.

DenisGabriela:
@montgolfiere, do you want to take another look on this before I merge it?

PR #55, opened 2019-06-11
REPO-4331: Internal release

eknizat:


.travis.yml
```diff
@@ -25,7 +25,7 @@ branches:
     - release/SP/2.0.N
     - company_release
     - /^ATS-.*$/
-  - /^feature.*$/
+    - /^feature.*$/
```
CezarLeahu: This list needs to include the `release_repo_4331` branch (as part of this PR).
Also, the list of branches for the *Release* job needs to include `release_repo_4331`, also - something like:
>(line 53)
```
if: branch IN (release, release/SP/1.3.N, release/SP/2.0.N, release_repo_4331)
```

CezarLeahu: Nvm, I see these changes have already been added on the `release_repo_4331` branch. No need to add them again.

CezarLeahu:
I suggest before the merge we change the next release version to `2.1.0-CUSTOM2`.
For this, you can use:
```bash
mvn versions:set -DnewVersion=2.1.0-CUSTOM2-SNAPSHOT versions:commit
```

PR #59, opened 2019-06-20
ATS-443 : Dependency updates

CezarLeahu:


pom.xml
```diff
@@ -18,7 +18,7 @@
         <java.version>1.8</java.version>
         <image.tag>latest</image.tag>
         <dependency.pdfbox.version>2.0.15</dependency.pdfbox.version>
-        <dependency.alfresco-core.version>7.3</dependency.alfresco-core.version>
+        <dependency.alfresco-core.version>7.14</dependency.alfresco-core.version>
         <dependency.alfresco-data-model.version>8.8</dependency.alfresco-data-model.version>
```
montgolfiere: Update alfresco-data-model (that is compatible with alfresco-core) ?

pom.xml
```diff
@@ -76,11 +76,31 @@
                 <version>${dependency.alfresco-transform-model.version}</version>
             </dependency>
             <!-- via Data Model / OpenCMIS Client -->
+            <dependency>
+                <groupId>org.apache.cxf</groupId>
+                <artifactId>cxf-core</artifactId>
+                <version>3.3.2</version>
```
montgolfiere: Should have single dependency version referenced (so that 3.3.2 can be changed in one place next time around) ?

pom.xml
CezarLeahu: I've updated both dependencies based on the latest ACS packaging pom.

pom.xml
CezarLeahu: Yes, we should. Done.
I've also downgraded the version to **3.2.9**, so it would pass our whitesource check during the build.

PR #60, opened 2019-06-24
ATS-467 : Instantiate JMS beans only when 'activemq.url' property is set

CezarLeahu:


alandavis:
Cezar Leahu:

The two modes (*JMS+HTTP* and *only HTTP*) are mutually exclusive. By not providing the ACTIVEMQ_URL variable, the T-Engines run in *only HTTP* implicitly (but spam the logs with error messages).
I’ve added a pull request that basically just implements your suggestion from the ATS-467 description:
> T-Engines should not be using Active MQ in the Community edition, so should not provide an ACTIVEMQ_URL as they do in the Enterprise edition. Perhaps the absence of this value should indicate that the connection is not required.
PR: https://github.com/Alfresco/alfresco-transform-core/pull/60

This changes should work well with the deployment referenced in the Story’s descrition: https://github.com/Alfresco/acs-community-packaging/blob/feature/LocalTransform/docker-alfresco/test/docker-compose.yml
#60 ATS-467 : Instantiate JMS beans only when 'activemq.url' property is set
Alfresco/alfresco-transform-core | Today at 11:59 AM | Added by GitHub
Instead of using a dedicated flag, the T-Engines would just check if the `ACTIVEMQ_URL` variable is set.

Alan Davis:
They are not mutually exclusive in the Enterprise Edition. We can have both types of request taking place to the same docker image. One from the Transform Service and the other from Local Transformers in the repo. We need both in order to allow customers to migrate their own custom transformers into T-Engines. Sometime a customer will need to combine in a local pipeline their own custom T-Engine with another T-Engine such as the ImageMagick one. They should not have to turn off the Transform Service to do that or have duplicate T-Engine images.

This was not a problem about 10 days ago. Can we please go back to being able to use both types of connection in the same image.

montgolfiere:
Just to clarify JMS+HTTP (~ Enterprise) vs HTTP (~ Community). 

Future ACS Enterprise 6.2 (with future Transform Service 1.1) will:

* use async JMS (via T-Router) for remote transforms - if configured/enabled & transform is supported
* fallback to sync HTTP (direct to T-Engines) for local transform - if configured/enabled & transform is supported
* fallback to embedded transformers (deprecated) - if neither of the above cannot support the requested transform

Future ACS Community 6.2 will:
* use sync HTTP (direct to T-Engines) for local transformers - if configured/enabled & transform is supported
* fallback to embedded transformers (deprecated) - if the above cannot support the requested transform

PR #66, opened 2019-07-01
REPO-4331: Update target file size check

eknizat:
Transformers should not throw an exception when source and target files are empty


montgolfiere:
@eknizat Is this issue limited to Misc T-Engine ? I wonder if we also need a more general fix (possibly a different issue) in other T-Engines &/or Enterprise T-Router (cc @CezarLeahu).

alandavis:
if (sourceFile.length() > 0 && targetFile.length() == 0)

@montgolfiere, I'm not sure this is something you want as a general check as there could well be cases where a transforms might result in a zero length file. However in this case we needed to allow zero length source (Strings) to result in a zero length target. The problem was that the zero length result was seen as an error.

eknizat:
@montgolfiere This issue is limited to the Misc T-Engine from what I have seen so far. As Alan mentioned, it is caused by zero length source (Strings) transformations which get requested by solr as ACS starts up. I have not seen transformations like that for other T-Engines.

montgolfiere:
Great, thanks

eknizat:
Test Added.

PR #70, opened 2019-07-08
ATS-481: temp workaround for transitive dependency

montgolfiere:
- temp workaround attempt to update from dom4j 1.6.1 to 2.1.1
- transitive dependency (alfresco-core <- spring-surf-core-configservice)
- workaround should be removed once REPO-4514 resolved (future alfresco-core <- spring-surf)


montgolfiere:
I suggest mark this as done and then open a new task (minor "tech debt") to remind us to remove the workaround in the future. We can link the new task back to this one, as a reminder.

PR #71, opened 2019-07-08
ATS-467 : Disable JMS Config beans on missing "activemq.url" property

CezarLeahu:


CezarLeahu:
I'm merging the PR so we can move the task back into _Verify_.
The build is red only due to the WhiteSource scans (check https://github.com/Alfresco/alfresco-transform-core/pull/70)

PR #72, opened 2019-07-09
ATS-467 : JMS config startup message

CezarLeahu:


alfresco-transformer-base/src/main/java/org/alfresco/transformer/messaging/MessagingInfo.java
```diff
@@ -0,0 +1,30 @@
+package org.alfresco.transformer.messaging;
```
montgolfiere: Please also add licence header (as per other Java files), thanks.

CezarLeahu: Yes, I forgot about that (new IDE, new settings).
Done.

PR #78, opened 2019-07-17
MNT-20337/REPO-4534/ATS-484 LibreOffice 6.1.6

alandavis:
Can I have your 3 story points for ATS-484 :) Needed to work out if 6.1.6 was compatible with alfresco-jodconverter-core-3.0.1.1 which is used by both the Libreoffice T-Engine and the repo.


README.md
```diff
@@ -16,7 +16,7 @@ Contains the common transformer (T-Engine) code, plus a few actual implementatio
 
 The project can be built by running the Maven command:
 ```bash
-mvn clean install -Plocal
+mvn clean install -DskipTests -Plocal
```
alandavis: Requires ActiveMQ to be running and configured at the moment, so you might like to work out what needs to change in the README.md, before closing ATS-484.

alfresco-docker-libreoffice/Dockerfile
```diff
@@ -18,8 +18,8 @@ COPY target/alfresco-docker-libreoffice-${env.project_version}.jar libreoffice-d
 
 RUN ln /alfresco-docker-libreoffice-${env.project_version}.jar /usr/bin/alfresco-docker-libreoffice.jar && \
     yum install -y cairo cups-libs libSM && \
-    test -f libreoffice-dist-5.4.6-linux.gz && \
-    ln -s libreoffice-dist-5.4.6-linux.gz libreoffice-dist-linux.gz || \
+    test -f libreoffice-dist-6.1.6-linux.gz && \
```
alandavis: If I owned this code I would not have the version number in the code. Currently it needs to be updated in 3 places in this file and also in the executor (if the minor version changes).

montgolfiere:
Yes, of course - all tested contributions are gratefully accepted :-)

Do we know if the unit tests (in this project &/or Repo are sufficient to catch blatant regressions ?

README.md
montgolfiere: Yes, we should update the README steps (as needed) so that we do not skip tests when building locally. It might be as a separate PR.

alfresco-docker-libreoffice/Dockerfile
CezarLeahu: Done. I've added Dockerfile arguments for the _LibreOffice_ & _ImageMagick_ versions.

README.md
CezarLeahu: I've updated the line with the proper profiles.

CezarLeahu:
I've merged the **`master`** branch and also addressed the PR comments.
However, the build will fail at the WhiteSource scan due the the `quartz` dependency ([CVE-2019-13990](https://github.com/quartz-scheduler/quartz/issues/467)).

We'll most likely merge this PR once we run a few minor sanity tests.

PR #95, opened 2019-08-21
REPO-4617 Add missing source types for T-engines

Epurashu:
Working on [REPO-4452](https://issues.alfresco.com/jira/browse/REPO-4452) - Add local transformer pipelines, an investigation regarding the supported target->source supported mimetypes was done, please see [comment](https://issues.alfresco.com/jira/browse/REPO-4452?focusedCommentId=631901&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-631901)

```
Local: Number of source to target mimetype transforms: 75
Legacy: Number of source to target mimetype transforms: 3658

Local: Number of source to target mimetype transforms via renditions: 31
Legacy: Number of source to target mimetype transforms via renditions: 226
```
On the current number of supported source->target transform, the possibilities for the pipelines transformer are few.
This PR contains all the missing Source->Target mimetypes for each T-engine.

PR #100, opened 2019-08-27
Sync REPO-4617_missing_mimetypes with master

Epurashu:


montgolfiere:
How does this relate to https://github.com/Alfresco/alfresco-transform-core/pull/95 ??

Epurashu:
@montgolfiere yes, I need to sync with master because of the *TransformerIT tests classes.

montgolfiere:
So do still need https://github.com/Alfresco/alfresco-transform-core/pull/95 or can we close that one (without merge) ?

Epurashu:
The changes from #95 are not into master. This sync with master is to bring the integration tests classes but the tests will be part of #95

PR #108, opened 2019-09-04
WIP: Fix/repo 4639 content conversion tika

alandavis:
The Tika T-Engine "transform" option does not exist when called via the Transform Service or Local transforms, which resulted in no transforms taking place. However this value is really not be needed as the T-Engine should be able to read its own engine_config.xml to work out which sub transform to use. Transforms only worked via Legacy transforms, which used a T-Engine.

This code is based on tried and tested ACS repository code. It has been further simplified.

TODO:
- replace the ConfigFileFinder class just added with something that uses Spring to read the JSON. i.e. simplify it.
- replace the CombinedConfig class just added with something that does not need the InLineTransformer. i.e. simplify it.
- create tests based on the repo tests
- remove the source and target mimetype checks in Tika as a check against engine_config.xml is cleaner.
- repeat the process for the Misc T-Engine as it has similar code checking source and target mimetypes.
- remove the transform option passed by the legacy transforms.

### DIFF
--- a/alfresco-transformer-base/src/main/java/org/alfresco/transformer/AbstractTransformerController.java
+++ b/alfresco-transformer-base/src/main/java/org/alfresco/transformer/AbstractTransformerController.java
@@ -43,11 +43,15 @@ import static org.springframework.util.StringUtils.getFilenameExtension;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
 
 import org.alfresco.transform.client.model.TransformReply;
 import org.alfresco.transform.client.model.TransformRequest;
 import org.alfresco.transform.client.model.TransformRequestValidator;
 import org.alfresco.transform.client.model.config.TransformConfig;
+import org.alfresco.transform.client.model.config.TransformRegistry;
 import org.alfresco.transform.exceptions.TransformException;
 import org.alfresco.transformer.clients.AlfrescoSharedFileStoreClient;
 import org.alfresco.transformer.logging.LogEntry;
@@ -115,6 +119,9 @@ public abstract class AbstractTransformerController implements TransformControll
 
     @Autowired
     private ObjectMapper objectMapper;
+
+    @Autowired
+    private TransformRegistry transformRegistry;
 
     @GetMapping(value = "/transform/config")
     public ResponseEntity<TransformConfig> info()
@@ -351,5 +358,38 @@ public abstract class AbstractTransformerController implements TransformControll
         }
 
         return sb.toString();
+    }
+
+    protected String getTransformerName(File sourceFile, String sourceMimetype, String targetMimetype,
+                                      Map<String, String> transformOptions)
+    {
+        long sourceSizeInBytes = sourceFile.length();
+        String transformerName = transformRegistry.getTransformerName(sourceMimetype, sourceSizeInBytes,
+                targetMimetype, transformOptions, null);
+        if (transformerName == null)
+        {
+            throw new TransformException(BAD_REQUEST.value(), "No transforms were able to handle the request");
+        }
+        return transformerName;
+    }
+
+    protected Map<String, String> createTransformOptions(Object... namesAndValues)
+    {
+        if (namesAndValues.length % 2 != 0)
+        {
+            logger.error("Incorrect number of parameters. Should have an even number as they are names and values.");
+        }
+
+        Map<String, String> transformOptions = new HashMap<>();
+        for (int i=0; i<namesAndValues.length; i+=2)
+        {
+            String name = namesAndValues[i].toString();
+            Object value = namesAndValues[i + 1];
+            if (value != null)
+            {
+                transformOptions.put(name, value.toString());
+            }
+        }
+        return transformOptions;
     }
 }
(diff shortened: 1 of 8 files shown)


alfresco-transformer-base/src/main/java/org/alfresco/transformer/AbstractTransformerController.java
```diff
@@ -352,4 +336,37 @@ private static String messageWithCause(final String prefix, Throwable e)
...
+        String transformerName = transformRegistry.getTransformerName(sourceMimetype, sourceSizeInBytes,
+                targetMimetype, transformOptions, null);
+        if (transformerName == null)
+        {
+            throw new TransformException(BAD_REQUEST.value(), "No transforms were able to handle the request");
+        }
+        return transformerName;
+    }
+
+    protected Map<String, String> createTransformOptions(Object... namesAndValues)
```
CezarLeahu: Replace `Object...` with `Map<String, Object>`.

alfresco-docker-tika/src/test/java/org/alfresco/transformer/TikaControllerTest.java
```diff
@@ -300,6 +300,9 @@ public void dotDotSourceFilenameTest() throws Exception
         super.dotDotSourceFilenameTest();
     }
 
+
+//  Ignore test as source/target extension si mandatory parameter for Tika.
```
tzclucian: Are there any planned changes that I might not be aware of or can this test be deleted instead of ignored if it's not needed anymore?

alandavis: Thank you - will be removed.

alfresco-transformer-base/src/main/java/org/alfresco/transformer/AbstractTransformerController.java
alandavis: That would be nice, but also needs to not include keys to null values.
I need to check again, but it appears that alfresco-transform-core is Java 8 rather than >= 9, so can't use Map.of either and then remove the nulls later.

alfresco-docker-transform-misc/src/main/java/org/alfresco/transformer/MiscController.java
```diff
@@ -109,24 +108,27 @@ public void processTransform(final File sourceFile, final File targetFile,
...
     @PostMapping(value = "/transform", consumes = MULTIPART_FORM_DATA_VALUE)
     public ResponseEntity<Resource> transform(HttpServletRequest request,
         @RequestParam("file") MultipartFile sourceMultipartFile,
         @RequestParam("targetExtension") String targetExtension,
         @RequestParam("targetMimetype") String targetMimetype,
+        @RequestParam(value = "targetEncoding", required = false) String targetEncoding,
         @RequestParam("sourceMimetype") String sourceMimetype,
-        @RequestParam(value = "testDelay", required = false) Long testDelay,
-        @RequestParam Map<String, String> parameters)
+        @RequestParam(value = "sourceEncoding", required = false) String sourceEncoding,
```
eknizat: I would prefer to keep the parameters like pageLimit out of the MiscController and only specify parameters common for all transformers, and if the creation of the transformOptions could be done in a more generic way so that addition of other transformers to the Misc T-engine will not require unnecessary changes to other classes.

alfresco-docker-transform-misc/src/main/java/org/alfresco/transformer/transformers/AppleIWorksContentTransformer.java
```diff
@@ -76,15 +76,6 @@
     //                (225 x 173) preview-web.jpg
     //                 (53 x  41) preview-micro.jpg
 
-    @Override
-    public boolean isTransformable(String sourceMimetype, String targetMimetype,
-        Map<String, String> parameters)
-    {
-        return MIMETYPE_IWORK_KEYNOTE.equals(sourceMimetype) ||
```
eknizat: I can't comment on the actual import statement above, but there are unused imports for the MIMETYPE_* fields left behind in the transformers (this one and others).

alfresco-transformer-base/src/main/java/org/alfresco/transformer/AbstractTransformerController.java
montgolfiere: Should we agree to switch to minimum Java Source = 11 ? 

Also aligns with ACS Repo (https://github.com/Alfresco/alfresco-repository/blob/master/pom.xml#L38) and AIS.

alfresco-docker-alfresco-pdf-renderer/src/main/resources/engine_config.json
```diff
@@ -12,8 +12,8 @@
     {
       "transformerName": "pdfrenderer",
       "supportedSourceAndTargetList": [
-        {"sourceMediaType": "application/pdf",             "targetMediaType": "image/png" },
-        {"sourceMediaType": "application/illustrator",     "targetMediaType": "image/png" }
+        {"sourceMediaType": "application/pdf",            "priority": 50, "targetMediaType": "image/png" },
```
montgolfiere: We should also document the new "priority" feature (ie. behaviour) in the Transform Config API design notes.

alfresco-transformer-base/src/main/java/org/alfresco/transformer/AbstractTransformerController.java
CezarLeahu: I understand the need for this method now. Please disregard this comment/conversation.

PR #188, opened 2020-03-09
ATS-468 : Add Veracode (SAST & SCA) scans to Transform Service Travis builds

DenisGabriela:


.travis.yml
```diff
@@ -43,6 +43,14 @@ jobs:
       before_install: bash _ci/init.sh
       script: bash _ci/whitesource.sh
 
+    - name: "Source Clear Scan"
+      stage: build
+      if: branch NOT IN (company_release)
+      before_install: bash _ci/init.sh
+      script: skip
```
montgolfiere: Can we document somewhere (JIRA, Confluence, README ?) how this is configured ?  For example, policies that will cause the build to fail ?

DenisGabriela: I've added a few details here. Thanks for review.

tzclucian:
> We should not delete WhiteSource yet - not until we're fully up-to-speed with VeraCode.

It was deleted upon [this chat](https://alfresco.slack.com/archives/CT2NQUYE4/p1584438966008200?thread_ts=1584374888.003800&cid=CT2NQUYE4) on the #eng-security-veracode channel.
As I'm seing two sides on this topic ... which one should we follow? Is there any decision/documentation regarding this at an organizational level that we should stick to?.

cc: @mteodori

DenisGabriela:
Can I have one more approve here as well? Similar changes on transform-service have been approved&merged.

PR #196, opened 2020-03-30
ATS-692 Fix VSCode compiler error with Tika-parsers

PunchedChimera:
Fixes compiler/editor error: The package javax.xml.<artifactId> is accessible from more than one module.
Currently this is not showing in java compilers due to a bug in the compiler: https://bugs.openjdk.java.net/browse/JDK-8215739
Once this bug is fixed and the error is reported correctly this will not build correctly. Excluding xml-apis:xml-apis from the classpath will fix this.


PunchedChimera:
> Was this a recently introduced issue - eg. could it be fallout from recent ATS-674 ?

It appears so, this error is not present on SP/1.3N or SP/2.0.N.

alfresco-transform-tika/alfresco-transform-tika/pom.xml
```diff
@@ -49,6 +49,10 @@
                     <groupId>org.quartz-scheduler</groupId>
                     <artifactId>quartz</artifactId>
                 </exclusion>
+                <exclusion>
```
montgolfiere: Why is this change required now (but not previously) ? Are these libraries provided elsewhere ? If not, when does Tika use these libraries at runtime ?

montgolfiere:
Please also check 21.1 tag (in addition to SP/1.3N or SP/2.0.N).

PR #197, opened 2020-03-30
ATS-691: Make path to test files OS agnostic

KikoUK:


alfresco-transformer-base/src/test/java/org/alfresco/transformer/AbstractTransformerControllerTest.java
```diff
@@ -161,14 +162,26 @@ void generateTargetFileFromResourceFile(String actualTargetExtension, File testF
...
-            throw new IOException("The test file " + testFilename +
-                                  " does not exist in the resources directory");
+            File testFileUrl = null;
+            try {
+                testFileUrl = new File(new File(getClass().getProtectionDomain().getCodeSource().getLocation().toURI()) + "\\" + testFilename);
+            } catch (URISyntaxException e) {
+                e.printStackTrace();
+            }
+            return testFileUrl;
+        }
```
PunchedChimera: Is there need to add a separate implementation for windows? Would this new implementation not work on linux too?

KikoUK: Unfortunately it doesn't. Initially I also thought it would (see the first commit) but all tests on travis (runs linux VMs) failed with incorrect pathToFile.

PunchedChimera: It looks like the cause of your error on travis was building the the end of the file path manually, specifically the use of a backslash. Would converting the URL in the original code to a URI not have the same affect without having to add multiple implementations?
`183             return testFileUrl == null ? null : new File(testFileUrl.toURI())`
I'm sure some additional logic may be required to catch the `URISyntaxException`.

PunchedChimera: Also, in your current implementation, missing a file will produce a different error on windows as you only handle the `URISyntaxException`. A file not existing is specifically handled in the original code.

KikoUK: I explored that option yesterday, but you can't surround a return statement in try/catch and if you add the exception to the method signature, you have to add the exception to all calling methods in the stack as well, which didn't seem like a good option. What you suggest wouldn't work without changing the original code a lot and I chose to split the logic. 
Your second point, the windows exception is different than linux, but I tested that scenario with no file and it seems like it's throwing the correct exception when a file is missing. I guess is either because both exception inherit from Exception or because IOException is already part of the method's signature.

alfresco-transformer-base/src/test/java/org/alfresco/transformer/AbstractTransformerControllerTest.java
```diff
@@ -162,17 +162,26 @@ void generateTargetFileFromResourceFile(String actualTargetExtension, File testF
 
     protected File getTestFile(String testFilename, boolean required) throws IOException
     {
+        if (System.getProperty("os.name").substring(0, 3).equals("Win"))
+        {
             File testFileUrl = null;
             try {
```
PunchedChimera, line 168: Curly braces should be on new line and each part of try-catch should be on a new line : https://hub.alfresco.com/t5/alfresco-content-services-hub/coding-standards-for-alfresco-content-services/ba-p/290457

alfresco-transformer-base/src/test/java/org/alfresco/transformer/AbstractTransformerControllerTest.java
KikoUK: Ran through the suggested fixes, but they broke linux paths.

alfresco-transformer-base/src/test/java/org/alfresco/transformer/AbstractTransformerControllerTest.java
```diff
@@ -162,17 +162,26 @@ void generateTargetFileFromResourceFile(String actualTargetExtension, File testF
...
                 testFileUrl = new File(new File(getClass().getProtectionDomain().getCodeSource().getLocation().toURI()) + "\\" + testFilename);
```
PunchedChimera, line 169: In windows implementation the original error handling is not executed. When a file is not found a different error is being thrown for windows and is causing test failures.

KikoUK: Found the test failures. It wasn't actually due to the error handling or the line above, it was due to the logic the files are generated through FileChannel which expects null if the file doesn't exist for files which isRequired is set to false, which in the original code is handled by `return testFileUrl == null ? null : new File(testFileUrl.getFile());` 
I also added the exception from the original code for completeness (Although I think that's already handled)

alfresco-transformer-base/src/test/java/org/alfresco/transformer/AbstractTransformerControllerTest.java
```diff
@@ -161,14 +162,34 @@ void generateTargetFileFromResourceFile(String actualTargetExtension, File testF
...
+        {
+            ClassLoader classLoader = getClass().getClassLoader();
+            URL testFileUrl = classLoader.getResource(testFilename);
+            if (required && testFileUrl == null) {
+                throw new IOException("The test file " + testFilename +
+                        " does not exist in the resources directory");
+            }
+            return testFileUrl == null ? null : new File(testFileUrl.getFile());
         }
-        return testFileUrl == null ? null : new File(testFileUrl.getFile());
```
montgolfiere, line 171: Can we restructure slightly to have one return of testFileUrl ?

PR #201, opened 2020-04-02
REPO-4919 - LibreOffice 6.3.5

evasques:


montgolfiere:
Thanks @evasques.

Can you please update this PR from master *and* you will also need to update the extra new Dockerfile, see:

https://github.com/Alfresco/alfresco-transform-core/blob/master/alfresco-transform-core-aio/alfresco-transform-core-aio-boot/Dockerfile#L17

PR #202, opened 2020-04-02
ATS-695 AIO T-Engine Spring Boot

PunchedChimera:
Created a spring-boot app for the new All in One T-Engine.
Including:
AIOController
Application
transformForm

Test class has been added but still need implementing as future transform adapters are required.
Added test resources in preparation. (Majority of files in the PR. Any file with the name "quick")
edit: Test resources have now been removed from PR, these are now planned to be added when tests are implemented.


PunchedChimera:
Build will fail as controller is expecting classes from the target branch

montgolfiere:
@PunchedChimera thanks !

Please add a short description to PR to summarise the list of changes (or use commit messages as bullets) ...

I see 84 files sounds like a lot of changes - is that because test classes have been duplicated. If so, no worries, but maybe raise JIRA tech debt (on component backlog) for future review & rationalisation ?

montgolfiere:
I'm also clearing the W/S scan (ATS-577). Someone still needs to check the V/C scan issue ...

PunchedChimera:
Thanks @montgolfiere I've added a description. The majority of the files added were test resources in preparation for testing. As the test cannot be implemented without the transform adapters being written I will add a specific JIRA ticket. Would you prefer these test resources be brought in as part of this new ticket?

PunchedChimera:
The V/C scan has not been amended to include the new jar created by the new aio-boot app. There are often issues with concurrent scan being executed by multiple Travis builds. I'll take a look at it.

PR #214, opened 2020-04-09
ATS-676 - test release of T-Core (T-Engines) 2.2.0-TEST1 [trigger release]

montgolfiere:
- bump pom.xml to 2.2.0-TEST1-SNAPSHOT
- update time_wait from 30 to 45 mins (since 2.2.0-A1 may have timed-out)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,7 @@ jobs:
       if: commit_message ~= /\[trigger release\]/ AND branch ~= /^(master|SP\/.+|HF\/.+)$/
       before_install: bash _ci/init.sh
       before_script: travis_wait bash _ci/cache_artifacts.sh
-      script: travis_wait 45 bash _ci/release.sh
+      script: travis_wait 55 bash _ci/release.sh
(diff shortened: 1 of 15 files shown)


.travis.yml
```diff
@@ -63,7 +63,7 @@ jobs:
       if: commit_message ~= /\[trigger release\]/ AND branch ~= /^(master|SP\/.+|HF\/.+)$/
       before_install: bash _ci/init.sh
       before_script: travis_wait bash _ci/cache_artifacts.sh
-      script: travis_wait 30 bash _ci/release.sh
+      script: travis_wait 45 bash _ci/release.sh
```
eknizat, line 66: I would go even higher just in case, maybe an hour

montgolfiere: OK, I'll bump to 55 and we can come back down later, if possible.

PunchedChimera:
The release stage -dryrun is not building the docker images. As this was the point of failure on master do we want to include this?

montgolfiere:
I'm not sure - on master it seemed to timeout (it is building & pushing 1 extra docker image now).

PR #218, opened 2020-04-14
ATS-714: Fix libreoffice NPE when debug enabled

eknizat:
Fix NPE in libreoffice when the external files are not present and debug is enabled.


montgolfiere:
FYI: I've updated to master (which should also fix/remove the broken SAST job).

PR #228, opened 2020-04-20
ATS-707: add transformName override

eknizat:
* Provide a way for Legacy ACS transformers to specify the transform to perform, rather than letting the T-Engine to determine it using the engine config.


alfresco-transform-core-aio/alfresco-transform-core-aio-boot/src/main/java/org/alfresco/transformer/AIOController.java
```diff
@@ -71,6 +71,10 @@
 {
     private static final Logger logger = LoggerFactory.getLogger(AIOController.class);
 
+    // This property can be sent by acs repository's legacy transformers to force a transform,
+    // instead of letting this T-Engine determine it based on the request parameters
+    private static final String TRANSFORM_NAME_PROPERTY = "transformName";
```
montgolfiere, line 76: Can you give examples of names ? 

Also, mark as deprecated ?

eknizat: should I mark the private variable with @Depreceated?

PR #229, opened 2020-04-20
ATS-728: Add optional Windows parameters for ImageMagickCommandExecutor

KikoUK:


alfresco-transform-imagemagick/alfresco-transform-imagemagick/src/main/java/org/alfresco/transformer/executors/ImageMagickCommandExecutor.java
```diff
@@ -76,6 +80,14 @@ protected RuntimeExec createTransformCommand()
...
         processProperties.put("DYLD_FALLBACK_LIBRARY_PATH", DYN);
         processProperties.put("LD_LIBRARY_PATH", DYN);
+
+        //Optional Windows properties
+        if ((CODERS != null && !CODERS.isBlank())
+                && (CONFIG != null && !CONFIG.isBlank()))
+        {
+            processProperties.put("MAGICK_CODER_MODULE_PATH", CODERS);
+            processProperties.put("MAGICK_CONFIGURE_PATH", CONFIG);
+        }
```
montgolfiere, line 90: Are these two properties always mutually dependent or can then be set independently ?

If the latter, then we should split into two separate if blocks ?

KikoUK: I assumed all 5 properties are required (and all 5 are different than the linux ones) for Windows base on this repo props file. [link](https://github.com/Alfresco/alfresco-repository/blob/d1216de8d447280952f5a4c32885144a33ddbb89/src/main/resources/alfresco/subsystems/thirdparty/default/imagemagick-transform.properties). I couldn't find any relevant ImageMagick docs which prove that. I am currently trying to test that in Windows environment.

KikoUK: Decided to split for simplicity and Alfresco [docs ](https://docs.alfresco.com/6.2/tasks/imagemagick-config.html)already specify what should be included for windows. Since all 5 props need to be different than default linux props, the permutations of user errors are too many to catch in this case.

alfresco-transform-imagemagick/alfresco-transform-imagemagick/src/main/java/org/alfresco/transformer/executors/ImageMagickCommandExecutor.java
```diff
@@ -76,6 +80,16 @@ protected RuntimeExec createTransformCommand()
         processProperties.put("MAGICK_HOME", ROOT);
         processProperties.put("DYLD_FALLBACK_LIBRARY_PATH", DYN);
         processProperties.put("LD_LIBRARY_PATH", DYN);
+
+        //Optional Windows properties
```
montgolfiere, line 84: Add https://imagemagick.org/script/resources.php#environment link to comment ?

alfresco-transform-imagemagick/alfresco-transform-imagemagick/src/main/java/org/alfresco/transformer/executors/ImageMagickCommandExecutor.java
```diff
@@ -76,6 +80,16 @@ protected RuntimeExec createTransformCommand()
...
+
+        //Optional properties (see also https://imagemagick.org/script/resources.php#environment)
+        if (CODERS != null && !CODERS.isBlank())
+        {
+            processProperties.put("MAGICK_CODER_MODULE_PATH", CODERS);
+        }
+        if (CONFIG != null && !CONFIG.isBlank())
+        {
+            processProperties.put("MAGICK_CONFIGURE_PATH", CONFIG);
+        }
```
PunchedChimera, line 92: Do both these have to be set in order to work? If they each work on their own then don't worry.
Do we want some error handling if this is the case?  
if (CODERS ==null AND CONFIG!=null){throw error("you need to supply both CODERS and CONFIG)}

KikoUK: Please refer to this [comment ](https://github.com/Alfresco/alfresco-transform-core/pull/229#pullrequestreview-396635253) where we discussed the above with Jan

PR #203, opened 2020-04-06
ATS-669: Parameterize T-Engines transformer execution locations

KikoUK:

PR #231, opened 2020-04-21
ATS-702: Add AIO tests from LibreOffice

PunchedChimera:
Runs the LibreOffice transformer tests against the new All in One transformer


alfresco-transform-libreoffice/alfresco-transform-libreoffice-boot/src/main/resources/application-default.yaml
```diff
@@ -5,4 +5,4 @@ transform:
     config:
       location: classpath:libreoffice_engine_config.json
     libreoffice:
-      home: ${LIBREOFFICE_HOME:/opt/libreoffice6.3}
```
eknizat, line 8: Should this be changing?

alfresco-transform-libreoffice/alfresco-transform-libreoffice-boot/src/main/java/org/alfresco/transformer/LibreOfficeController.java
```diff
@@ -79,7 +79,7 @@
 {
     private static final Logger logger = LoggerFactory.getLogger(LibreOfficeController.class);
 
-    @Value("${transform.core.libreoffice.home}")
```
eknizat, line 82: intentional?

alfresco-transform-core-aio/alfresco-transform-core-aio-boot/src/test/java/org/alfresco/transformer/AIOControllerLibreOfficeTest.java
```diff
@@ -0,0 +1,124 @@
...
+        adapter = new LibreOfficeAdapter(execPath);
+    }
+
+    @Override
+    // Used by the super class to mock the javaExecutor, a different implementation is required here
+    protected void setJavaExecutor(AbstractTransformerController controller, LibreOfficeJavaExecutor javaExecutor)
+    {
+        ReflectionTestUtils.setField(adapter, "javaExecutor", javaExecutor);
+        //Need to wire in the mocked adapter into the controller...
+        if (ReflectionTestUtils.getField(transformRegistry,"transformerTransformMapping") instanceof Map)
```
eknizat, line 75: What's the purpose of this if statement? If transformerTransformMapping is not a Map this test should be failing anyway.

eknizat: You can replace the whole thing with just:
Map<String,Transformer> transformers = transformRegistry.getTransformerTransformMapping();
transformers.replace("libreoffice", adapter);
No need to override it with reflection after either.

alfresco-transform-libreoffice/alfresco-transform-libreoffice-boot/src/main/java/org/alfresco/transformer/LibreOfficeController.java
PunchedChimera: Yes, both the AIO controller and the LibreOffice controller uses an application.yml file to define this variable. As they were given different variable names it prevented inheritance of the test class.

eknizat: Ah yes they should be the same but right now in aio application.yml it is libreoffice.path and in libreoffice it is libreoffice.home. @KikoUK any one in particular it should be?

alfresco-transform-core-aio/alfresco-transform-core-aio-boot/src/test/java/org/alfresco/transformer/AIOControllerLibreOfficeTest.java
PunchedChimera: Yes you're right, there was originally some type safety issues in a older version that is no longer present. Removed the additional lines.
Also made the change in ImageMagick

alfresco-transform-libreoffice/alfresco-transform-libreoffice-boot/src/main/resources/application-default.yaml
PunchedChimera: Yes, both the AIO controller and the LibreOffice controller application.yml file were given different variable names, which prevented inheritance of the test class.

alfresco-transform-libreoffice/alfresco-transform-libreoffice-boot/src/main/java/org/alfresco/transformer/LibreOfficeController.java
KikoUK: I tried to keep it consistent as requested by Jan in these comments: https://github.com/Alfresco/alfresco-transform-core/pull/203#discussion_r408943587
https://github.com/Alfresco/alfresco-transform-core/pull/203#discussion_r408944665
I kept the original name of the var OFFICE_HOME everywhere, but then there were a few requests to change it to path. I agree that it needs to be the same in both aio and libreoffice, I guess we need to agree on which name to use. I vote for OFFICE_HOME as the method which sets it is called setOfficeHome. So ..office.home

PR #243, opened 2020-05-15
ATS-735: Update to ImageMagick 7.0.10-11 distro bins

TheOutsSeeker:
Updated IMAGEMAGICK_VERSION reference in both Dockerfiles


TheOutsSeeker:
As discussed with the team. Branch and PR build are good.

PR #246, opened 2020-05-20
ATS-731 Update T-Engines config with remaining legacy transformers

KikoUK:
Draft PR to run travis tests


alfresco-transform-core-aio/alfresco-transform-core-aio-boot/src/main/java/org/alfresco/transformer/AIOController.java
```diff
@@ -135,11 +127,7 @@ protected void executeTransformCommand(File sourceFile, File targetFile)
         @RequestParam(SOURCE_MIMETYPE) String sourceMimetype,
         @RequestParam(TARGET_MIMETYPE) String targetMimetype,
         @RequestParam Map<String, String> requestParameters,
-        @RequestParam (value = TEST_DELAY, required = false) Long testDelay,
-
-        // The TRANSFORM_NAME_PROPERTY param allows ACS legacy transformers to specify which transform to use,
-        // It can be removed once legacy transformers are removed from ACS.
```
montgolfiere, line 141: Is this comment line true ? Do we actually need to wait for legacy transformers to actually be removed from ACS ... ?

... or, as per this PR, remove now since we are adding T-Config for all remaining transforms (~ legacy) that are supported by the current T-Engines.

In other words, I want to ensure that even if we remove this code, release a new version of T-Engines (eg. 2.2.2) *and* deploy with existing ACS 6.2.1 then it will continue to work via local transforms as per new T-Config.

cc @eknizat @PunchedChimera

montgolfiere:
Early comment (I realise PR is still Draft) ;-)

Please check with @PunchedChimera whether it is ok to remove duplicate T-Config json files from "boot" directories, eg. for imagemagick & libreoffice.

montgolfiere:
@KikoUK @aymanthefirst What's left to do to make this PR available for review and merge ?

There are also a couple of early review comments above which would be good to resolve.

alfresco-transform-core-aio/alfresco-transform-core-aio-boot/src/main/java/org/alfresco/transformer/AIOController.java
montgolfiere: @KikoUK What's the plan to resolve above comment ? 

If we decide to leave above code in (until we know new transforms are working, eg. via automated tests) then we should also add an INFO log message to indicate if Repo is still unexpectedly going via this code path.

montgolfiere:
> Draft PR to run travis tests

Please update first comment. PR is no longer draft - also which Travis tests ?

PR #247, opened 2020-05-25
REPO-4334 Move metadata extraction into T-Engines

alandavis:
- Metadata extract code added to T-Engines
- Required a refactor of duplicate code to avoid 3x more duplication:
  - try catches use to return return exit codes
  - calls to java libraries or commands to external processes
  - building of transform options in controllers, adaptors
- integration tests based on current extracts performed in the repo
- included extract code for libreoffice, and embed code even though not used out of the box.
  There may well be custom extracts are still using it and will need the code move to T-Engines

PR #256, opened 2020-06-12
ATS-777 / REPO-4334: Move metadata extraction into T-Engines

alandavis:
- Added a default implementation of AbstractTransformerController#transform(java.lang.String, java.lang.String, java.lang.String, java.util.Map<java.lang.String,java.lang.String>, java.io.File, java.io.File) for existing
T-Engines that override transform(HttpServletRequest...) and processTransform(...).

I would get rid of the code duplication in alfresco-ai-transformers. There is some rework required in its' test class mocking anyway. From the log: "There is already 'aiController' bean method ... transform(". The test code appear to assume there can be only be one method called transform. This is no longer the case. I'm guessing it is a small fix for someone with the AWS dev env set up, so it might make sense to fix up the code duplication at the same time.

### DIFF
--- a/alfresco-transform-core-aio/alfresco-transform-core-aio-boot/src/main/java/org/alfresco/transformer/AIOController.java
+++ b/alfresco-transform-core-aio/alfresco-transform-core-aio-boot/src/main/java/org/alfresco/transformer/AIOController.java
@@ -82,7 +82,7 @@ public class AIOController extends AbstractTransformerController
             {
                 Map<String, String> parameters = new HashMap<>();
                 parameters.put(SOURCE_ENCODING, "UTF-8");
-                transform("html", MIMETYPE_HTML, MIMETYPE_TEXT_PLAIN, parameters, sourceFile, targetFile);
+                processTransform("html", MIMETYPE_HTML, MIMETYPE_TEXT_PLAIN, parameters, sourceFile, targetFile, null);
             }
         };
     }
@@ -95,9 +95,17 @@ public class AIOController extends AbstractTransformerController
         return new ResponseEntity<>(transformConfig, OK);
     }
 
+    @Deprecated
+    public void processTransform(final File sourceFile, final File targetFile,
+                                 final String sourceMimetype, final String targetMimetype,
+                                 final Map<String, String> transformOptions, final Long timeout)
+    {
+        processTransform(null, sourceMimetype, targetMimetype, transformOptions, sourceFile, targetFile, timeout);
+    }
+
     @Override
-    protected void transform(String transformName, String sourceMimetype, String targetMimetype,
-                             Map<String, String> transformOptions, File sourceFile, File targetFile)
+    public void processTransform(String transformName, String sourceMimetype, String targetMimetype,
+                                    Map<String, String> transformOptions, File sourceFile, File targetFile, Long timeout)
     {
         logger.debug("Processing transform with: transformName; '{}', sourceFile '{}', targetFile '{}', transformOptions" +
                 " {}", transformName, sourceFile, targetFile, transformOptions);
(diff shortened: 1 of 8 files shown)


montgolfiere:
@alandavis sorry, I'm confused by your description ? 

Is this PR needed or not ? How does it relate to the original https://github.com/Alfresco/alfresco-transform-core/pull/247 (for REPO-4334) ?

Are both PRs backwards compatible in terms of custom T-Engines ?

alandavis:
Jan, what you are calling deprecated methods in sub classes of AbstractTransformerController still exist and if implemented in custom transforms will be called. So not deprecated.

The problem is there is a new abstract transform method, which should now ideally be implemented. This PR now provides a blank implementation so that custom transforms will compile and still work. If custom transforms still provide the original methods (the ones you are calling deprecated) the new abstract method is not called.

Qu: so why not do what is provided in this PR, to get the code custom T-Engine code to work?
Ans: because ideally it should be abstract to point developers at the correct way to implement the code without adding duplicate code.

If you would rather have full backward Java compatibility merge the PR. If you would rather have a design that makes the creation of new T-Engines simpler, rework the original ones by adding a blank method, or even remove the duplicate code.

montgolfiere:
I'm not sure this PR is the solution. 

However, I think the new method should also include "timeout" which seems to be missing here :-(

https://github.com/Alfresco/alfresco-transform-core/blob/master/alfresco-transformer-base/src/main/java/org/alfresco/transformer/AbstractTransformerController.java#L406-L419

The existing documentation refers to "processTransform" so we need updates to the README to show how a T-Engine can transition - ie. handling/ignoring additional parameters such as "transformName" and "timeout" ?

https://github.com/Alfresco/acs-packaging/blob/master/docs/creating-a-t-engine.md#processtransform

montgolfiere:
@alandavis I've tried an alternative which also adds missing timeout parameter.

alandavis:
I can see what you are trying to do.
Timeout - I think there is a bigger problem, that has nothing to do with this rework. Southhall bug? No transform threads are being terminated by this value, other than possibly libreoffice.
In the http transform case, you have been passing null in as the new timeout property. The timeout value is actually part of the transformOptions, so is likely to be ignored.
Thought: Could we not do the same with the original processTransform method. i.e. pass in the timeout as one of the transformOptions. Not sure why your old team would have separated it anyway. This might also avoid having to change the signature.
As we currently don’t have many custom T-Engines, we probably still have the opportunity to refactor without introducing deprecated methods. Overriding the old processTransform in all T-Engines does not feel good. I don’t think it is worth it.
processTransform vs transform method names: I don’t really care other than that I’m updating docs with transform at the moment.
Happy to talk this over.

montgolfiere:
I'm not sure how the revert helps ? 

We're still left with:

Currently documented interface which no longer needs to be implemented

https://github.com/Alfresco/alfresco-transform-core/blob/2.2.3/alfresco-transformer-base/src/main/java/org/alfresco/transformer/TransformController.java#L68-L70

Implementation of current interface that does not set timeout (if needed):

https://github.com/Alfresco/alfresco-transform-core/blob/2.2.3/alfresco-transformer-base/src/main/java/org/alfresco/transformer/AbstractTransformerController.java#L406-L419

New method that needs to be implemented but not part of the interface:

https://github.com/Alfresco/alfresco-transform-core/blob/2.2.3/alfresco-transformer-base/src/main/java/org/alfresco/transformer/AbstractTransformerController.java#L487-L488

Lack of JavaDoc makes it harder. Either way, we also need this documentation to be updated:

https://github.com/Alfresco/acs-packaging/blob/master/docs/creating-a-t-engine.md#processtransform

montgolfiere:
@alandavis  and I have updated this PR. The new method to implement is "transformImpl(...)". 

Alan mentioned that he has a separate JIRA to update the Markdown and hello-world sample transformer, including:

- https://github.com/Alfresco/acs-packaging/blob/master/docs/creating-a-t-engine.md#processtransform

- https://github.com/Alfresco/alfresco-helloworld-transformer

On that basis, I have ticked 👍

montgolfiere:
cc @eknizat @PunchedChimera @KikoUK .. ok to merge ?

PR #257, opened 2020-06-12
MNT-21487: Switch from smaller schemas jar (poi-ooxml-schemas) to the full jar (ooxml-schemas)

amohammedalfresco:
- Apply changes of #255 to branch *SP/2.1.N*


amohammedalfresco:
Seems WhiteSourceScan failure is regardless of the changes and also in other branches.

montgolfiere:
@amohammedalfresco Do we need this at this stage since I'm not aware of immediate plans for 2.1.N ? If not, we can close without merge for now.

Note: We encourage upgrade to 2.2.N which should be backwards compatible.

montgolfiere:
As per comment above, I've closed without merge for now.

PR #266, opened 2020-06-25
ATS-738 / MNT-21613: Allow option to use -DACTIVEMQ_URL (without needing to also set -Dactivemq.url)

montgolfiere:
- note: in the meantime, workaround(s) exist including simply using ACTIVEMQ_URL env var

### DIFF
--- a/alfresco-transformer-base/src/main/java/org/alfresco/transformer/messaging/MessagingInfo.java
+++ b/alfresco-transformer-base/src/main/java/org/alfresco/transformer/messaging/MessagingInfo.java
@@ -2,7 +2,7 @@
  * #%L
  * Alfresco Transform Core
  * %%
- * Copyright (C) 2005 - 2019 Alfresco Software Limited
+ * Copyright (C) 2005 - 2020 Alfresco Software Limited
  * %%
  * This file is part of the Alfresco software.
  * -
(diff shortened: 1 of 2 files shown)


alfresco-transformer-base/src/main/java/org/alfresco/transformer/messaging/MessagingInfo.java
```diff
@@ -49,8 +49,19 @@
...
+        // Note: as per application.yaml the broker url is appended with "?jms.watchTopicAdvisories=false". If this needs to be fully
+        // overridden then it would require explicitly setting both "spring.activemq.broker-url" *and* "activemq.url" (latter to non-false value).
+
+        if ((activemqUrl != null) && (! activemqUrl.equals("false")))
+        {
+            logger.info("JMS client is ENABLED - ACTIVEMQ_URL ='{}'", activemqUrl);
+        }
+        else
+        {
+            logger.info("JMS client is DISABLED - ACTIVEMQ_URL is not set");
```
KikoUK, line 64: Shouldn't this info log include the current workaround on how to set the url as a temporary message?

alfresco-transformer-base/src/main/resources/application.yaml
```diff
@@ -14,6 +14,9 @@ spring:
   jackson:
     default-property-inclusion: non_empty
 
+activemq:
+  url: ${ACTIVEMQ_URL:false}
```
KikoUK, line 18: Can we set default values for both ActiveMQ?

montgolfiere: There is default value (= false) if ACTIVEMQ_URL is not set.

I'm not sure how else we can be backwards compatible *and* support the ability to conditionally configure JMS (rather than http-only) as done via:

@ConditionalOnProperty(name = "activemq.url")

PR #330, opened 2021-02-04
Include throwable cause in TransformException

marsv024:
When an exception occurs in transformations we will get the message and and internal server error. However the cause is lost. In my case I got a NullPointerException in the core-aio transform, but no info where it originated from. With this change the cause will be added to the new TransformException.

### DIFF
--- a/alfresco-transformer-base/src/main/java/org/alfresco/transformer/executors/Transformer.java
+++ b/alfresco-transformer-base/src/main/java/org/alfresco/transformer/executors/Transformer.java
@@ -76,11 +76,11 @@ public interface Transformer
         }
         catch (IllegalArgumentException e)
         {
-            throw new TransformException(BAD_REQUEST.value(), getMessage(e));
+            throw new TransformException(BAD_REQUEST.value(), getMessage(e), e);
         }
         catch (Exception e)
         {
-            throw new TransformException(INTERNAL_SERVER_ERROR.value(), getMessage(e));
+            throw new TransformException(INTERNAL_SERVER_ERROR.value(), getMessage(e), e);
         }
         if (!targetFile.exists())
         {


montgolfiere (approved):
Thanks @marsv024 for the contribution 👍 

LGTM. I notice we already do it in a few other places (where we catch and rethrow the exception).

PunchedChimera (approved):

KikoUK (approved):

eknizat (approved):

montgolfiere:
Pending green build.

TheOutsSeeker:
@montgolfiere 

We won't get a green build here ever, please see https://docs.travis-ci.com/user/pull-requests#pull-requests-and-security-restrictions

`The way encrypted environment variables work is that you use the travis gem to add encrypted variables to your .travis.yml. Those keys are then automatically decoded by Travis at build time, but only for builds triggered by contributors with write access to the repository — not for forks, meaning not for external contributors’ PRs.`

References describing the issue:
- https://github.com/OHDSI/Achilles/issues/165 (PR 17 Nov 2016)
- https://www.algolia.com/blog/engineering/travis-encrypted-variables-external-contributions/ (Feb 13th 2018)

Solution: 
Take the suggested code change, raise an internal PR, pending internal approvals and green build, merge into master and finally update and close this PR in order to inform external contributor.

montgolfiere:
@TheOutsSeeker What we do for our other public projects, eg alfresco-community-repo & many others ?

TheOutsSeeker:
> @TheOutsSeeker What we do for our other public projects, eg alfresco-community-repo & many others ?

I don't know, perhaps something to be clarified within the circle of the software architects. My best guess is, that those projects are public ones hence can be checked out directly by external contributor and PRs raised without forking the project as such.

PR #367, opened 2021-04-07
MNT-22082 transformation of pdf to text hang

PunchedChimera:
See [MNT-22082](https://alfresco.atlassian.net/browse/MNT-22082)
The hanging is caused only when `notExtractBookmarkText` is set to false.
A workaround was added to repo for local transforms only and was not applied to ATS. This workaround set the default value of `notExtractBookmarkText` to true.  As of 7.0, local transforms have been removed. This PR will now add a similar workaround for both the Tika and AIO T-engines, and by extension ATS.

A new constructor has been added to the `TikaController` to provide the new spring config. The original constructor has been maintained so that there are no issue with backward compatibility.
The creation of the `TikaJavaExecutor` has been moved to "singleton pattern" as the injection of the `@Value` happens after the instantiation of the `TikaJavaExecutor` and does not pass the value correctly, if done as a member variable in the previous manner. The instantiation is now done once, on the first transform request.
Param has been added to the AIO beans.

### DIFF
--- a/alfresco-transform-tika/alfresco-transform-tika/src/main/java/org/alfresco/transformer/executors/TikaJavaExecutor.java
+++ b/alfresco-transform-tika/alfresco-transform-tika/src/main/java/org/alfresco/transformer/executors/TikaJavaExecutor.java
@@ -118,9 +118,9 @@ public class TikaJavaExecutor implements JavaExecutor
         final boolean includeContents = parseBoolean(
                 transformOptions.getOrDefault(RequestParamMap.INCLUDE_CONTENTS, "false"));
         final boolean notExtractBookmarksText = parseBoolean(
-                transformOptions.getOrDefault(NOT_EXTRACT_BOOKMARKS_TEXT, String.valueOf(notExtractBookmarksTextDefault)));
+                transformOptions.getOrDefault(RequestParamMap.NOT_EXTRACT_BOOKMARKS_TEXT, String.valueOf(notExtractBookmarksTextDefault)));
         final String targetEncoding = transformOptions.getOrDefault("targetEncoding", "UTF-8");
-        if(transformOptions.get(NOT_EXTRACT_BOOKMARKS_TEXT)==null && notExtractBookmarksTextDefault)
+        if(transformOptions.get(RequestParamMap.NOT_EXTRACT_BOOKMARKS_TEXT)==null && notExtractBookmarksTextDefault)
         {
             LoggerFactory.getLogger(TikaJavaExecutor.class).trace(
                     "notExtractBookmarksText default value has been overridden to {}",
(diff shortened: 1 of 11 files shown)


montgolfiere:
Thanks @PunchedChimera 

FYI: see also https://github.com/Alfresco/alfresco-transform-core/pull/361

Out of interest, is the failure also repeatable with more recent versions of PDFBox (after 2.0.22) ?

https://downloads.apache.org/pdfbox/
- 2.0.23 ( 2021-03-18 ) 
- 3.0.0-RC1 ( 2021-04-01 )

For the latter, we should raise tech debt reminder to consider upgrade to 3.0.0 in the future when it is generally-available (can compatible with Tika etc).

alfresco-transform-tika/alfresco-transform-tika/src/main/java/org/alfresco/transformer/executors/TikaJavaExecutor.java
```diff
@@ -109,9 +118,14 @@ public void transform(String transformName, String sourceMimetype, String target
         final boolean includeContents = parseBoolean(
                 transformOptions.getOrDefault(RequestParamMap.INCLUDE_CONTENTS, "false"));
         final boolean notExtractBookmarksText = parseBoolean(
-                transformOptions.getOrDefault(NOT_EXTRACT_BOOKMARK_TEXT, "false"));
+                transformOptions.getOrDefault(NOT_EXTRACT_BOOKMARK_TEXT, String.valueOf(notExtractBookmarksTextDefault)));
```
montgolfiere, line 121: It might be out-of-scope but I wonder if we should consider renaming "NOT_EXTRACT_BOOKMARK_TEXT" to "NOT_EXTRACT_BOOKMARKS_TEXT" ?

https://github.com/Alfresco/alfresco-transform-core/blob/master/alfresco-transformer-base/src/main/java/org/alfresco/transformer/util/RequestParamMap.java#L73

This would then also be consistent with:

https://github.com/Alfresco/alfresco-transform-core/blob/master/alfresco-transform-tika/alfresco-transform-tika/src/main/java/org/alfresco/transformer/executors/Tika.java#L464

`public static final String NOT_EXTRACT_BOOKMARKS_TEXT = "--notExtractBookmarksText";`

PunchedChimera:
> Thanks @PunchedChimera
> 
> FYI: see also #361
> 
> Out of interest, is the failure also repeatable with more recent versions of PDFBox (after 2.0.22) ?
> 
> https://downloads.apache.org/pdfbox/
> 
> * 2.0.23 ( 2021-03-18 )
> * 3.0.0-RC1 ( 2021-04-01 )
> 
> For the latter, we should raise tech debt reminder to consider upgrade to 3.0.0 in the future when it is generally-available (and compatible with Tika etc).

2.0.23 doesn't fix the issue and 3.0.0-RC1 doesn't work at all.

montgolfiere (approved):

KikoUK (approved):

PR #414, opened 2021-05-26
ATS-913: Tech debt: add more T-Core Unit tests (part 2) for IPTC metadata extract

aymanthefirst:
Added RAW tests and their corresponding files for IPTC metadata extraction.
Including: 
File type | MB
-- | --
ARW | 23.8
RW2 | 11.2
CR2 | 18.4
NEF | 22.8
RAF | 32.1
Total | 108.3


montgolfiere (changes requested):
As per ATS-913, what about other existing quick image formats ?

eg. quickIPTC.png, quickIPTC.gif ... etc (maybe consistent with quickIPTC.jpg or similar)

For the RAW files, what is the source of the original files please ?

aymanthefirst:
@montgolfiere I just saw your last comment and change of description on https://alfresco.atlassian.net/browse/ATS-913

Now I understand the problem.

So I will make make a new PR with those changes. It shouldn't take long as the time-consuming part of building the project is done. 

What shall we do with this PR?

KikoUK (approved):

PunchedChimera (approved):

PR #415, opened 2021-05-27
Refactor TikaController

PunchedChimera:
Allows for object creation to be done on startup, instead of first transform call.


montgolfiere:
@PunchedChimera Still needed ? If so, I suggest we also log in JIRA for tracking (reason etc).

mpichura (approved):
Looks ok to me.

PunchedChimera:
@montgolfiere I see no reason to not merge it. I don't see any need to track it (we could just consider it training/ iTime)
It was a small refactor, does not affect functionality in any way.

montgolfiere:
Thanks

montgolfiere (approved):

PR #428, opened 2021-06-15
ACS-1677 Provide file extension from JMS request if missing in SFS

PunchedChimera:
This will provide the file downloaded from Share File Store with the expected file extension provided in the `TransformRequest`, instead of null, if it is not present in the Share File Store response.


KikoUK (approved):

montgolfiere (commented):
Do this impact (fix ?) any tests (eg. in this or other related projects) ?

PunchedChimera:
All of the tests in this project are http requests, so this will not affect any test in this project. It may enable some test files in alfresco-transform-service, that are currently disable to be re-enabled.

montgolfiere (approved):

PR #468, opened 2021-09-15
WIP: ATS-968 switch ATS Core java base image to CentOS 7

mpichura:
ImageMagick is built from source to maintain the same version as used with CentOS 8.  
Project builds without errors but wasn't yet tested for regression.  
Still needs replacing ImageMagick download page with Nexus one.


pzhyland:
Can be closed. Upgrade will be provided in different PRs

PR #469, opened 2021-09-21
ATS-970: Fix build after Travis secrets rotation

montgolfiere:
- use QUAY_USERNAME  (set to "alfresco+travis_ci")
- (instead of old hard-coded "alfresco+bamboo" robot account that no longer exists)
- update settings.xml to use env vars for usernames (consistent with other projects)


KikoUK (approved):

PunchedChimera (approved):
Subject to green build

PR #471, opened 2021-09-24
ACS-2003 Tika transformers will not create temporary files on disk

levbelava:
Source file will not be created anymore for Tika transformer.

All changes are backwards compatible for all transformers.
All APIs which used temp file on disk is workable (Tika including).
Old processing path is workable and callable in Tika too (see TikaController for example).
Similar work can be done for Misc transformers.

Probe Test in Tika  transformers now is configurable and can use old temp file on disk path or not.


montgolfiere:
FYI :ACS-2003 was closed as an experiment. Hence, closing this Draft for now.

PR #484, opened 2021-11-11
ACS-1998 Enhance the transform config

alandavis:
* Use CombinedTransformConfig from the transform-model rather than trying to combine TransformConfig in this project


montgolfiere:
@alandavis Merged without review ?

montgolfiere:
I see you've also released T-Core/ (T-Engines) 2.5.4. I've added a short description to release:

https://github.com/Alfresco/alfresco-transform-core/releases/tag/2.5.4

PR #507, opened 2022-01-02
ACS-2002 Enhance T-Router debug

alandavis:
By default T-Engines now only provide TransformerDebug DEBUG messages, rather than the original detailed request and reply messages, which are still available as TRACE.


alfresco-transformer-base/src/main/java/org/alfresco/transformer/config/WebApplicationConfig.java
```diff
@@ -78,4 +79,10 @@ public TransformServiceRegistry transformRegistry()
     {
         return new TransformRegistryImpl();
     }
```
alandavis, line 81: We need to avoid adding debug to the clientData from both t-router and t-engines. This flag is used.

pzhyland (approved):

PR #512, opened 2022-01-06
ACS-2382 License reporting.

tpage-alfresco:


_ci/prepare_release_deploy.sh
```diff
@@ -10,11 +10,9 @@ export VERSION=$(git describe --abbrev=0 --tags)
 
 mkdir -p deploy_dir
 
-# Download the WhiteSource report
```
montgolfiere, line 13: Minor: although unrelated, maybe also take the opportunity to remove other whitesource related config (or else in defer to separate PR) eg.

- https://github.com/Alfresco/alfresco-transform-core/blob/master/.travis.yml#L80-L94
- https://github.com/Alfresco/alfresco-transform-core/blob/master/.whitesource
- https://github.com/Alfresco/alfresco-transform-core/blob/master/_ci/whitesource.sh

montgolfiere (approved):
Added minor suggestion

alandavis (approved):
Local build also looks like it will do the release job fine now.

PunchedChimera (approved):

PR #516, opened 2022-01-11
Repeat Bump dependency.tika.version from 2.1.0 to 2.2.1

alandavis:
Original PR https://github.com/Alfresco/alfresco-transform-core/pull/506 was merged to master where it failed. There had been no build of the PR before the merge, which is why this branch has been created.

With the upgrade of Tika from 2.2.0 to 2.2.1, the deprecated OfficeOpenXMLCore.SUBJECT metadata value became  null and the replacement TikaCoreProperties.SUBJECT became a multi value in a few of our test cases. For backward compatibility with very old versions of Alfresco, we have historically been added a number of extra values including "subject" and "description" back into the raw metadata, before mapping them onto Alfresco properties. These values existed in the original version of Tika used by Alfresco, so it is possible there are custom mappings out there that using them.

To complicate matters a little, out standard mappings for some types put the raw "subject" value into the cm:description property. What makes it interesting is that the extra "description" value is not used but has the value originally in our expected metadata expected data. This is why the quick_*_json files have been modified.

### DIFF
--- a/alfresco-transform-tika/alfresco-transform-tika-boot/src/test/resources/quick.odf_metadata.json
+++ b/alfresco-transform-tika/alfresco-transform-tika-boot/src/test/resources/quick.odf_metadata.json
@@ -1,5 +1,5 @@
 {
-  "{http://www.alfresco.org/model/content/1.0}description" : "Gym class featuring a brown fox and lazy dog",
+  "{http://www.alfresco.org/model/content/1.0}description" : "Pangram, fox, dog, Gym class featuring a brown fox and lazy dog",
   "{http://www.alfresco.org/model/content/1.0}created" : 1138362922000,
   "{http://www.alfresco.org/model/content/1.0}author" : null,
   "{http://www.alfresco.org/model/content/1.0}title" : "The quick brown fox jumps over the lazy dog"
--- a/alfresco-transform-tika/alfresco-transform-tika/src/main/java/org/alfresco/transformer/metadataExtractors/AbstractTikaMetadataExtractor.java
+++ b/alfresco-transform-tika/alfresco-transform-tika/src/main/java/org/alfresco/transformer/metadataExtractors/AbstractTikaMetadataExtractor.java
@@ -269,7 +269,7 @@ public abstract class AbstractTikaMetadataExtractor extends AbstractMetadataExtr
 
             // Get the subject and description, despite things not
             //  being nearly as consistent as one might hope
-            String subject = getMetadataValue(metadata, OfficeOpenXMLCore.SUBJECT);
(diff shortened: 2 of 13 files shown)


alfresco-transformer-base/src/test/java/org/alfresco/transformer/TestFileInfo.java
```diff
@@ -76,4 +76,10 @@ public static TestFileInfo testFile(final String mimeType, final String extensio
     {
         return new TestFileInfo(mimeType, extension, path, false);
     }
+
```
alandavis, line 79: Makes it possible to see which test file is being used when run in Intelij

alfresco-transformer-base/src/test/java/org/alfresco/transformer/AbstractMetadataExtractsIT.java
```diff
@@ -102,8 +102,8 @@ public void testTransformation(TestFileInfo testFileInfo)
             jsonObjectMapper.writerWithDefaultPrettyPrinter().writeValue(actualMetadataFile, actualMetadata);
 
             Map<String, Serializable> expectedMetadata = readExpectedMetadata(metadataFilename, actualMetadataFile);
```
alandavis, line 104: Makes it possible to see which test file is being used when run on Travis

alfresco-transform-tika/alfresco-transform-tika/src/main/java/org/alfresco/transformer/metadataExtractors/AbstractTikaMetadataExtractor.java
```diff
@@ -269,30 +269,30 @@ private ParseContext buildParseContext(Metadata metadata, String sourceMimeType)
 
             // Get the subject and description, despite things not
             //  being nearly as consistent as one might hope
```
alandavis, line 271: The deprecated OfficeOpenXMLCore.SUBJECT is now null upgrading from Tika 2.2.0 to 2.2.1

alfresco-transform-tika/alfresco-transform-tika-boot/src/test/resources/quick.odf_metadata.json
```diff
@@ -1,5 +1,5 @@
 {
```
alandavis, line 1: Now that the raw metadata "subject" is multi line (two in these cases) there is a  "Pangram, fox, dog, " to the previously expected value.

alandavis:
Build https://app.travis-ci.com/github/Alfresco/alfresco-transform-core/builds/244769469 was green - not showing up in this PR.

PR #525, opened 2022-02-07
ACS-2497 T-Core: Accept DAU requests

kmagdziarz:
Added usage implementation of Direct Access URL.
Now Transform method will used DAU instead of multipart if provided.


alfresco-transformer-base/src/main/java/org/alfresco/transformer/AbstractTransformerController.java
```diff
@@ -148,12 +150,12 @@
         return new ResponseEntity<>(transformConfig, OK);
     }
```
alandavis, line 152: General note: Please try to avoid changing the current style by adding or removing spaces. If you are going to do that it needs to be in a separate PR so it is not all mixed up with functional changes.

alfresco-transformer-base/src/main/java/org/alfresco/transformer/util/RequestParamMap.java
```diff
@@ -72,4 +72,5 @@
     String INCLUDE_CONTENTS        = "includeContents";
     String NOT_EXTRACT_BOOKMARKS_TEXT = "notExtractBookmarksText";
     String PAGE_LIMIT              = "pageLimit";
```
alandavis, line 74: You will sen in my PR this static final really should be in the model project now.

alfresco-transformer-base/src/main/java/org/alfresco/transformer/clients/AlfrescoSharedFileStoreClient.java
```diff
@@ -2,7 +2,7 @@
  * #%L
```
alandavis, line 2: 1. I don't think there should be any changes in this file as it is to do with the SFS and your change is to do with DAUs which is an alternative. Yes there may be commonality between the two classes, but it would be better to have a separate class. It might (only might) make sense to have a common abstract parent or share a component.

alfresco-transformer-base/src/main/java/org/alfresco/transformer/clients/AlfrescoSharedFileStoreClient.java
```diff
@@ -90,12 +89,29 @@ public FileRefResponse saveFile(File file)
...
+     *
+     * @param directUrl Direct Access Url
+     * @return ResponseEntity<Resource>
+     */
+    public ResponseEntity<Resource> getContentViaDirectUrl(String directUrl)
+    {
+        try
+        {
+            return restTemplate.getForEntity(directUrl,
+                    org.springframework.core.io.Resource.class);
```
alandavis, line 113: Can you remove the fully qualified name here?

alfresco-transformer-base/src/main/java/org/alfresco/transformer/AbstractTransformerController.java
```diff
@@ -167,10 +169,24 @@
                     + "targetExtension: '{}', requestParameters: {}", sourceMimetype, targetMimetype, targetExtension, requestParameters);
         }
```
alandavis, line 171: You will need to add DIECT_URL to NON_TRANSFORM_OPTION_REQUEST_PARAMETERS on L129.

I know this is still a WIP, so you probably will be refactoring the code so the originalFilename is created so the createTargetFileName is common to both paths.

kmagdziarz: Oh, it have to be there, otherwise it will not work. I had to accidentally removed that during cleaning.

alandavis (changes requested):

alandavis (approved):
LGTM - I made a few format changes.

SaraAspery (approved):
LGTM!

PR #536, opened 2022-02-18
ATS-979 Spike alpine support for Tika

mteodori:
* add alpine profile to tika project to build with custom Dockerfile.alpine and tag alpine-latest
* add new job to travis for Tika alpine
* use docker-maven-plugin to run as fabric8 breaks on rancher desktop

Ref. ATS-979


mteodori:
closing in favour of a safer switch from CentOS 7 to Rocky Linux 8

PR #436, opened 2021-07-08
ATS-864: Switch from fabric8 to docker mvn plugin

TheOutsSeeker:
Replaced fabric8-maven-plugin with docker-maven-plugin and performed pom cleanup:
* moved configurations from the plugin to the execution level to avoid overlap
* removed duplicated execution of spring-boot:repackage

Ref. ATS-864

PR #545, opened 2022-03-01
ACS-2613 Restore missing images

pzhyland:


alandavis:
Piotr says the behavior was changed by https://alfresco.atlassian.net/browse/ATS-864

pzhyland:
To be precise, it was introduced by #436

PR #586, opened 2022-04-05
ACS-2718 Streamline ATS t-model/t-core

alandavis:
* Merge alfresco-transform-model project into alfresco-transform-core - WITHOUT the git history (to be sure we are not accidentally reveling any secrets - we don’t think there are any)  
* Simplify Maven dependencies and build project
* Unify release process of combined t-model and t-core

**PLEASE DO NOT REVIEW THE ADDED .java FILES** as these have just been copied from the alfresco-transform-model project. The header of the TransformException.java is the only file changed, as it referred to the Repository rather than the T-Model. So **THERE ARE JUST 7 FILES TO CHECK AND 2 DELETES.** 

- It was possible to remove the https://github.com/Alfresco/alfresco-transform-model/tree/master/_ci directory and https://github.com/Alfresco/alfresco-transform-model/blob/master/docs/build-and-release-101.md from the original project. 

- The static-analysis and white source files from alfresco-transform-core were removed as they are no longer used.

- There was little tidy up of the pom.xml files, other than adding a new module to the top level pom, removing <scm> & <distributionManagement> sections, setting the parent pom of the t-model and one version dependency.

- SNAPSHOT versions are still produced as there are down stream projects, such as alfresco-transform-service, alfresco-transform-aspose and alfresco-ai-transformers.

- The README.md and build-and-release.md files were modified to include the new module.


SaraAspery (approved):
LGTM

kcichonczyk (approved):

PR #658, opened 2022-08-17
[ACS-3434] libgs Removal

jottley:
- Update aio and imagemagick Dockerfile to remove libgs after install
- Remove imagemagick eps tests
- Remove eps mappings from imagemagick config
- Remove quick.eps


skopf (approved):
LGTM

PR #734, opened 2023-02-20
Feature/acs 4460 mtls support

kcichonczyk:


engines/base/src/main/java/org/alfresco/transform/base/config/MTLSConfig.java
```diff
@@ -0,0 +1,186 @@
...
+
+    @Value("${client.ssl.trust-store-password:}")
+    private char[] trustStorePassword;
+
+    @Value("${client.ssl.trust-store-type:}")
+    private String trustStoreType;
+
+    @Bean()
+    @Scope("prototype")
+    public WebClient.Builder clientBuilder() throws KeyStoreException, IOException, CertificateException, NoSuchAlgorithmException, UnrecoverableKeyException
```
mstrankowski, line 82: both client builder and rest template? Are we somehow already forced to use both by existing code that couldn't be easily changed to a single approach?

engines/base/src/main/java/org/alfresco/transform/base/config/MTLSConfig.java
```diff
@@ -0,0 +1,186 @@
...
+import java.io.InputStream;
+import java.security.KeyManagementException;
+import java.security.KeyStore;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.UnrecoverableKeyException;
+import java.security.cert.CertificateException;
+
+@Configuration
+public class MTLSConfig {
```
mstrankowski, line 60: Config is named MTLSConfig, but it also works for TLS from what I see in method names "isTlsOrMtlsConfigured", in that case I think it's better to stick to TLS in naming as the less specific one.

engines/base/src/main/java/org/alfresco/transform/base/config/MTLSConfig.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    private boolean isTlsOrMtlsConfigured()
+    {
+        return isTruststoreConfigured() || isKeystoreConfigured();
+    }
+
+    private boolean isTruststoreConfigured()
+    {
+        return trustStoreResource != null;
```
mstrankowski, line 110: I think it would be good to do a parameter validation here already. What if someone provides a keystore but doesn't provide a password? What if they provide both but not keystore type? That kind of thing would be best taken care of immediately during set up.

engines/base/src/main/java/org/alfresco/transform/base/config/MTLSConfig.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    private boolean isKeystoreConfigured()
+    {
+        return keyStoreResource != null;
```
mstrankowski, line 115: same as above

engines/base/src/main/java/org/alfresco/transform/base/config/MTLSConfig.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    private WebClient.Builder createWebClientBuilderWithSslContext() throws UnrecoverableKeyException, CertificateException, KeyStoreException, IOException, NoSuchAlgorithmException
+    {
+        SslContextBuilder sslContextBuilder = SslContextBuilder.forClient();
```
mstrankowski, line 120: Why does there need to be such a distinction between the client version and restTemplate version, if even restTemplate uses a "CloseableHttpClient" beneath?

mstrankowski: Also, I'd consider moving the sslContext generation to a separate method that does only that

engines/base/src/test/java/org/alfresco/transform/base/MtlsTestUtils.java
```diff
@@ -0,0 +1,69 @@
...
+public class MtlsTestUtils {
+
+    public static boolean isMtlsEnabled()
+    {
+        return Boolean.parseBoolean(System.getProperty("mtls-enabled"));
+    }
+
+    public static CloseableHttpClient httpClientWithMtls() throws NoSuchAlgorithmException, KeyManagementException, UnrecoverableKeyException, KeyStoreException, IOException, CertificateException {
+
+        String keyStoreFile = System.getProperty("mtls-keystore-file");
```
mstrankowski, line 32: "mtls-keystore-file" Where do these names come from? I don't really see them elsewhere in the code or in a properties file provided for tests. I'd expect "client.ssl...." or "server.ssl...." properties to be used

engines/base/src/test/java/org/alfresco/transform/base/clients/HttpClient.java
```diff
@@ -27,8 +28,8 @@
  */
 public class HttpClient
 {
-    private static final RestTemplate REST_TEMPLATE = new RestTemplate();
-    
+    private static final RestTemplate REST_TEMPLATE = MtlsTestUtils.isMtlsEnabled() ? MtlsTestUtils.restTemplateWithMtls() : new RestTemplate();
```
mstrankowski, line 31: I'd move that condition to inside ot MtlsTestUtils.
Call MtlsTestUtils.getRestTemplate() and in that method return either the mtls one or just RestTemplate

engines/base/src/test/java/org/alfresco/transform/base/clients/SfsClient.java
```diff
@@ -75,7 +76,7 @@ public static String uploadFile(final String fileToUploadName, final String sfsB
             .addPart("file", new FileBody(file, ContentType.DEFAULT_BINARY))
             .build());
 
-        try (CloseableHttpClient client = HttpClients.createDefault())
+        try (CloseableHttpClient client = MtlsTestUtils.isMtlsEnabled() ? MtlsTestUtils.httpClientWithMtls() : HttpClients.createDefault())
```
mstrankowski, line 79: Same as above, I'd move it to MtlsTestUtils

engines/base/src/test/java/org/alfresco/transform/base/clients/SfsClient.java
```diff
@@ -134,7 +135,8 @@ public static boolean checkFile(final String uuid, final String sfsBaseUrl) thro
             sfsBaseUrl+"/alfresco/api/-default-/private/sfs/versions/1/file/{0}",
             uuid));
 
-        try (CloseableHttpClient client = HttpClients.createDefault())
+        try (CloseableHttpClient client = MtlsTestUtils.isMtlsEnabled() ?
```
mstrankowski, line 138: Same as above

engines/base/src/test/java/org/alfresco/transform/base/MtlsTestUtils.java
```diff
@@ -0,0 +1,69 @@
...
+        SSLContext sslContext = sslContextBuilder.build();
+        SSLConnectionSocketFactory sslContextFactory = new SSLConnectionSocketFactory(sslContext);
+        return HttpClients.custom().setSSLSocketFactory(sslContextFactory).build();
+    }
+
+    public static RestTemplate restTemplateWithMtls()
+    {
+        ClientHttpRequestFactory requestFactory = null;
+        try {
+            requestFactory = new HttpComponentsClientHttpRequestFactory(httpClientWithMtls());
```
mstrankowski, line 63: Wait, so in tests the RestTemplate can be created using an HttpClient created with one version of initialising SSLContext and in config file it isn't that way? I want that config file to look similar to this, with 1 separate method responsible for creating an SslContext :D

kcichonczyk: Those are different classes from different dependencies - SSLContext vs SslContext, HttpClient from apache vs HttpClient from netty. HttpClient configured for test classes here is different than the one configured for WebClient - and it has different usage here.

engines/base/src/test/java/org/alfresco/transform/base/MtlsTestUtils.java
kcichonczyk: Those are for test use only - specified in the pom.xml file for Transform Service end-to-end test:
`<systemPropertyVariables>
      <mtls-enabled>true</mtls-enabled>
      <mtls-keystore-file>./enterprise/integration-test/keystore.p12</mtls-keystore-file>
      <mtls-keystore-type>PKCS12</mtls-keystore-type>
      <mtls-keystore-password>password</mtls-keystore-password>
      <mtls-truststore-file>./enterprise/integration-test/truststore.p12</mtls-truststore-file>
      <mtls-truststore-type>PKCS12</mtls-truststore-type>
      <mtls-truststore-password>password</mtls-truststore-password>
  </systemPropertyVariables>`

engines/base/src/main/java/org/alfresco/transform/base/config/MTLSConfig.java
kcichonczyk: Those are different classes from different dependencies - SSLContext vs SslContext, HttpClient from apache vs HttpClient from netty.

engines/base/src/main/java/org/alfresco/transform/base/config/MTLSConfig.java
kcichonczyk: I'm using both as this is how existing code was configured.

engines/base/src/main/java/org/alfresco/transform/base/config/MTLSConfig.java
kcichonczyk: I don't think that this requires another validation - it will fail anyway with proper exceptions like wrong format or wrong password.

engines/base/src/test/java/org/alfresco/transform/base/MtlsTestUtils.java
mstrankowski: I'd recommend moving the creation of the sslContext to a method called getApacheSslContext or something along the lines

mstrankowski (approved):

pzhyland (approved):

PR #770, opened 2023-03-29
[ACS-4460] add no hostname verification remove com.google.collections…

kcichonczyk:
… dependencies

### DIFF
--- a/engines/base/pom.xml
+++ b/engines/base/pom.xml
@@ -57,6 +57,11 @@
         </dependency>
 
         <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>org.dom4j</groupId>
             <artifactId>dom4j</artifactId>
         </dependency>
--- a/engines/base/pom.xml
+++ b/engines/base/pom.xml
@@ -79,11 +79,6 @@
             <artifactId>pooled-jms</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.google.collections</groupId>
-            <artifactId>google-collections</artifactId>
-            <version>1.0</version>
-        </dependency>
-        <dependency>
             <groupId>ch.qos.logback</groupId>
             <artifactId>logback-classic</artifactId>
         </dependency>
--- a/engines/base/src/main/java/org/alfresco/transform/base/config/MTLSConfig.java
+++ b/engines/base/src/main/java/org/alfresco/transform/base/config/MTLSConfig.java
@@ -28,9 +28,12 @@ package org.alfresco.transform.base.config;
 
 import io.netty.handler.ssl.SslContext;
 import io.netty.handler.ssl.SslContextBuilder;
+import io.netty.handler.ssl.util.InsecureTrustManagerFactory;
 import org.alfresco.transform.base.WebClientBuilderAdjuster;
+import org.apache.http.conn.ssl.NoopHostnameVerifier;
 import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
 import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.impl.client.HttpClients;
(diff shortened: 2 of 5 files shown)


engines/base/pom.xml
```diff
@@ -56,6 +56,11 @@
             </exclusions>
         </dependency>
 
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
+
```
mstrankowski, line 63: Why the switch from collections to guava?

pzhyland (approved):

engines/base/src/main/java/org/alfresco/transform/base/config/MTLSConfig.java
```diff
@@ -164,7 +173,10 @@ private HttpClient createHttpClientWithSslContext(SslContextBuilder sslContextBu
     private RestTemplate createRestTemplateWithSslContext(SSLContextBuilder sslContextBuilder) throws NoSuchAlgorithmException, KeyManagementException {
         SSLContext sslContext = sslContextBuilder.build();
         SSLConnectionSocketFactory sslContextFactory = new SSLConnectionSocketFactory(sslContext);
-        CloseableHttpClient httpClient = HttpClients.custom().setSSLSocketFactory(sslContextFactory).build();
+
+        HttpClientBuilder httpClientBuilder = HttpClients.custom().setSSLSocketFactory(sslContextFactory);
+        if(hostNameVerificationDisabled) httpClientBuilder.setSSLHostnameVerifier(new NoopHostnameVerifier());
```
mstrankowski, line 178: That's great, but we also need to take a look at what Domenico provided - the security vulnerability

"If the Veracode finding actually makes sense, you guys could try customizing the Netty secure HTTP client creation similarly to how the reactor people do, and then agree with Ciro to mitigate the vulnerability as dealt with. E.g.:
    private HttpClient createHttpClientWithSslContext(SslContextBuilder sslContextBuilder) throws SSLException {
        SslContext sslContext = sslContextBuilder.build();
        return HttpClient.create().secure(p -> p.sslContext(sslContext).handlerConfigurator(handler -> {
            SSLEngine engine = handler.engine();
            SSLParameters sslParameters = engine.getSSLParameters();
            sslParameters.setEndpointIdentificationAlgorithm("HTTPS");
            engine.setSSLParameters(sslParameters);
        }));
    }"
    
Which means that we should add a hostname verification manually by default when we use a Netty httpClient directly

engines/base/pom.xml
kcichonczyk: guava contains collections already, 1.0.0 was very old version, caused problems in other projects (ai-renditions)

engines/base/src/main/java/org/alfresco/transform/base/config/MTLSConfig.java
kcichonczyk: updated

engines/base/src/main/java/org/alfresco/transform/base/config/MTLSConfig.java
```diff
@@ -132,7 +140,10 @@ public SslContextBuilder nettySslContextBuilder() throws UnrecoverableKeyExcepti
             sslContextBuilder.keyManager(keyManagerFactory);
         }
 
-        if(isTruststoreConfigured())
+        if(hostNameVerificationDisabled)
+        {
+            sslContextBuilder.trustManager(InsecureTrustManagerFactory.INSTANCE);
```
mstrankowski, line 145: https://stackoverflow.com/questions/64190476/what-is-the-equivalent-of-setsslhostnameverifiernoophostnameverifier-instance
InsecureTrustManagerFactory doesn't restrict itself to hostnames only, it disregards all security.

mstrankowski: And can't you just add an if at the "createHttpClientWithSslContext" method then?

mstrankowski (approved):

PR #799, opened 2023-05-17
ACS-5191 Update to license-maven-plugin 2.0.1.

tpage-alfresco:


tpage-alfresco:
Including neater way to avoid running license header check in the model module.

mpichura (approved):

PR #824, opened 2023-06-13
MNT-23660: TIFF files fail to transform into PNG & JPG

krdabrowski:
- set startPage and endPage to 0 as default for NON multi-page image targets


tpage-alfresco (approved):
LGTM

mpichura (approved):

PR #843, opened 2023-07-31
Fix/acs 5039 libreoffice home arm64

wojtekswieton:


engines/libreoffice/Dockerfile
```diff
@@ -46,4 +46,4 @@ EXPOSE 8090
 
 USER ${LIBREUSERNAME}
 
-ENTRYPOINT java $JAVA_OPTS -jar /usr/bin/${env.project_artifactId}.jar
+ENTRYPOINT ["/bin/sh", "-c", "if [[ \"$(uname -m)\" != \"x86_64\" ]]; then export LIBREOFFICE_HOME=${LIBREOFFICE_HOME:=/usr/lib64/libreoffice}; fi && java $JAVA_OPTS -jar /usr/bin/${env.project_artifactId}.jar"]
```
gionn, line 49: better move to a dedicated script?

not sure why we are using ENTRYPOINT here which is usually useful when you package CLI utilies and you expects user be able to provide additional arguments while using `docker run`. If it's a server application CMD should be enough.

engines/aio/Dockerfile
```diff
@@ -79,4 +79,6 @@ EXPOSE 8090
 
 USER ${AIOUSERNAME}
 
-ENTRYPOINT java $JAVA_OPTS -jar /usr/bin/${env.project_artifactId}.jar
+# Copy the script to set LIBREOFFICE_HOME
+COPY docker-entrypoint.sh /docker-entrypoint.sh
+ENTRYPOINT ["/docker-entrypoint.sh", "${env.project_artifactId}"]
```
gionn, line 84: it's common practice to always end a file with an EOL, be consistent adding it

gionn (approved):

aonych (approved):

damianujma (approved):

dsibilio (approved):

PR #885, opened 2023-11-14
[MNT-23960] Added new options (pdfFont, pdfFontSize) to textToPdf transformer

tiagosalvado10:


jottley (approved):

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
```diff
@@ -231,22 +277,35 @@ static PDType1Font getStandardFont(String name)
         }
         //duplicating until here
 
+        private String fontName = null;
+        private boolean fontChanged = false;
```
tpage-alfresco, line 281: I don't think this is used.

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
```diff
@@ -344,6 +403,137 @@ public PDDocument createPDFFromText(Reader text, int pageLimit)
...
+
+            if (font != null)
+            {
+                super.setFont(font);
+            }
+
+            this.fontName = aFontName;
+        }
+
+        private PDFont getFont(PDDocument doc, String name)
```
tpage-alfresco, line 419: This method is quite complex. I'd suggest either splitting it up (maybe even extracting some logic to a new class) or adding javadoc to describe what it does.

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
```diff
@@ -344,6 +403,137 @@ public PDDocument createPDFFromText(Reader text, int pageLimit)
...
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+
+            if (font == null)
+            {
+                if (!name.equals(defaultFont))
```
tpage-alfresco, line 498: Is `defaultFont` the same as `getDefaultFont()` above? Is this the default font for the whole transformer or something specified by the PDF document?

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
```diff
@@ -344,6 +403,137 @@ public PDDocument createPDFFromText(Reader text, int pageLimit)
...
+        public String getFontName()
+        {
+            return this.fontName;
+        }
+
+        public String getDefaultFont()
+        {
+            if (defaultFont == null || defaultFont.isBlank())
+            {
+                return PDType1Font.HELVETICA.getName();
```
tpage-alfresco, line 520: In the config it looks like the default is NotoSans-Regular, but here it's Helvetica. Is there a reason for two different default fonts? Do we need to ship Helvetica in the docker image (as we're doing with NotoSans)?

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
```diff
@@ -78,19 +91,29 @@ public class TextToPdfContentTransformer implements CustomTransformerFileAdaptor
     private static final byte BB = (byte) 0xBB;
     private static final byte BF = (byte) 0xBF;
 
+    @Value("${transform.core.misc.pdfBox.defaultFont}")
```
dsibilio, line 94: This syntax usually fails if the property is not found, e.g.: a user overrides the property file and does not specify the `transform.core.misc.pdfBox.defaultFont` property. If that's the case, I think it could be sensible to provide a default on the code-level via the appropriate `@Value(property:default)` syntax as well, similarly to what we have done here: https://github.com/Alfresco/alfresco-transform-core/pull/885/files#diff-7b94454e8766efe02993468560e80b26b41e4dfeedb2361efef81b907cbe44beR29?

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
```diff
@@ -78,19 +91,29 @@ public class TextToPdfContentTransformer implements CustomTransformerFileAdaptor
     private static final byte BB = (byte) 0xBB;
     private static final byte BF = (byte) 0xBF;
 
+    @Value("${transform.core.misc.pdfBox.defaultFont}")
+    String pdfBoxDefaultFont;
```
dsibilio, line 95: Can't this be `private`?

engines/misc/src/test/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformerTest.java
```diff
@@ -43,14 +47,19 @@
...
+import org.apache.pdfbox.pdmodel.font.PDType1Font;
+import org.apache.pdfbox.text.PDFTextStripper;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;;
 
 public class TextToPdfContentTransformerTest
 {
     TextToPdfContentTransformer transformer = new TextToPdfContentTransformer();
 
+    private static final String TEXT_WITH_ABREVE = "Gămbardella, Matthew, Corets, Evă";
```
tpage-alfresco, line 60: I think this (and the next variable) should have an extra underscore in the name:
```suggestion
    private static final String TEXT_WITH_A_BREVE = "Gămbardella, Matthew, Corets, Evă";
```

engines/misc/src/test/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformerTest.java
```diff
@@ -146,6 +155,78 @@ public void testUTF8WithoutBOM() throws Exception
...
+
+        TransformCheckResult result = transformTextAndCheck(sourceFile, encoding, TEXT_WITHOUT_ABREVE, String.valueOf(-1), true,
+                parameters, false);
+
+        assertEquals(result.getUsedFont(), PDType1Font.TIMES_ROMAN.getName());
+        assertNull(result.getErrorMessage());
+    }
+
+    /**
+     * Test if a different font can be chosen to perform the transformation with abreve character. This test
```
tpage-alfresco, line 206: I think it's "breve" not "abreve": https://en.wikipedia.org/wiki/Breve
```suggestion
     * Test if a different font can be chosen to perform the transformation with breve character. This test
```

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
```diff
@@ -78,19 +91,29 @@ public class TextToPdfContentTransformer implements CustomTransformerFileAdaptor
     private static final byte BB = (byte) 0xBB;
     private static final byte BF = (byte) 0xBF;
 
+    @Value("${transform.core.misc.pdfBox.defaultFont}")
+    String pdfBoxDefaultFont;
 
     private final PagedTextToPDF transformer;
 
+    protected static final int DEFAULT_FONT_SIZE = 10;
```
dsibilio, line 99: Can't this be private? Also, I'd move it up next to the other `private static final` constants of this class as a best practice.

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
```diff
@@ -112,6 +135,11 @@ public void setFontSize(int fontSize)
         }
     }
 
+    public String getUsedFont()
+    {
+        return transformer.getFontName();
+    }
+
```
dsibilio, line 142: Is this used anywhere?

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
```diff
@@ -130,6 +158,24 @@ public void transform(final String sourceMimetype, final String targetMimetype,
...
+        Integer fontSize = null;
+        if (pdfFontSize != null)
+        {
+            try
+            {
+                fontSize = parseInt(pdfFontSize, PDF_FONT_SIZE);
+            }
+            catch (Exception e)
+            {
+                fontSize = DEFAULT_FONT_SIZE;
```
dsibilio, line 176: Swallowing the exception without logging anything, not even at `DEBUG` level, makes it harder for users and us to troubleshoot issues potentially related to non-parseable `pdfFontSize` values.

Also, if `pdfFontSize` is null we don't initialize the `fontSize` at all. Is this wanted?

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
```diff
@@ -231,22 +277,35 @@ static PDType1Font getStandardFont(String name)
...
             try
             {
+                doc = new PDDocument();
+
+                final PDFont font = getFont(doc, pdfFontName);
+                final int fontSize = pdfFontSize != null ? pdfFontSize : getFontSize();
+
+                fontName = font.getName();
+
+                logger.debug("Going to use font " + fontName + " with size " + fontSize);
```
dsibilio, line 301: We can use the [SLF4J placeholders](https://www.slf4j.org/manual.html#typical_usage) to format strings rather than concatenating them.

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
```diff
@@ -344,6 +403,137 @@ public PDDocument createPDFFromText(Reader text, int pageLimit)
...
+
+                if (mapping != null && mapping.getFont() != null && !mapping.isFallback())
+                {
+                    try
+                    {
+                        font = PDType0Font.load(doc, mapping.getFont().getOriginalData());
+                    }
+                    catch (Exception e)
+                    {
+                        logger.error("Error loading mapping font: " + e.getMessage());
```
dsibilio, line 446: We should proably log the `name` as well, unless it's already part of the exception message.

Also, SLF4J placeholders can be used again, instead of concatenating strings.

dsibilio (changes requested):

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
tiagosalvado10: You're correct, the `fontChanged`variable is a leftover from previous tests that have been removed.

engines/misc/src/test/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformerTest.java
tiagosalvado10: As the error fixed by this PR mentioned the `abreve` word I tended to use it across the code, however, I have changed the `abreve` occurrences to `breve`.

> java.lang.IllegalArgumentException: U+0103 ('abreve')

engines/misc/src/test/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformerTest.java
tiagosalvado10: As the error fixed by this PR mentioned the `abreve` word I tended to use it across the code, however, I have changed the `abreve` occurrences to `breve`.

> java.lang.IllegalArgumentException: U+0103 ('abreve')

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
tiagosalvado10: I have added the default value at code level as suggested so we have an additional safeguard.

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
tiagosalvado10: Changed `pdfBoxDefaultFont` visibility to `private`.

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
tiagosalvado10: Changed `DEFAULT_FONT_SIZE` visibility to `private` and moved its location up next to the other constants.

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
tiagosalvado10: It is only used for test purposes like `setStandardFont` and `setFontSize` methods.

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
tiagosalvado10: Added `name` to the log message and SLF4J placeholders instead string concatenation as suggested.

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
tiagosalvado10: The main class `TextToPdfContentTransformer` has 2 class level attributes:

`String pdfBoxDefaultFont`
- Used to store the value from `transform.core.misc.pdfBox.defaultFont` property ([here](https://github.com/Alfresco/alfresco-transform-core/blob/1fa2ed033007e7ade468e594e169b735c154a373/engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java#L94))

`PagedTextToPDF transformer`
- Responsible for performing the transformation
- The `PagedTextToPDF` class also has a `defaultFont` attribute ([here](https://github.com/Alfresco/alfresco-transform-core/blob/1fa2ed033007e7ade468e594e169b735c154a373/engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java#L282)) that is initialized [here](https://github.com/Alfresco/alfresco-transform-core/blob/1fa2ed033007e7ade468e594e169b735c154a373/engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java#L109) - this way the transformer is aware of the configured default font, otherwise, it wouldn't be able to get the value from Spring boot property.

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
tiagosalvado10: I have created a new `DEFAULT_FONT` constant containing `NotoSans-Regular` value so it can be used instead of `Helvetica`, even it's not expected to fallback in this validation, we can keep the consistency in the default font.

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
tiagosalvado10: Added log message in case the font size parsing fails. 

Even if `fontSize` is null, it will be supplied to the `createPDFFromText` method, where the value will be validated and initialized ([here](https://github.com/Alfresco/alfresco-transform-core/blob/1fa2ed033007e7ade468e594e169b735c154a373/engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java#L297)).

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
tiagosalvado10: Splitted `getFont(PDDocument, String)` method code into 3 new methods, each one with a different approach to get the font:

- `getFromStandardFonts(String fontName)`
- `getFromFontMapper(String, PDDocument)`
- `getFromFileSystem(String)`

Added Javadoc to each method.

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
```diff
@@ -130,6 +158,25 @@ public void transform(final String sourceMimetype, final String targetMimetype,
...
+        if (pdfFontSize != null && !pdfFontSize.isBlank())
+        {
+            try
+            {
+                fontSize = parseInt(pdfFontSize, PDF_FONT_SIZE);
+            }
+            catch (Exception e)
+            {
+                fontSize = DEFAULT_FONT_SIZE;
+                logger.error("Error parsing font size {}, going to set it as {} : {}", pdfFontSize, fontSize, e.getMessage());
```
tpage-alfresco, line 177: I think you should use standard slf4j exception logging for this:
```suggestion
                logger.error("Error parsing font size {}, going to set it as {}", pdfFontSize, fontSize, e);

```

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
```diff
@@ -344,6 +403,208 @@ public PDDocument createPDFFromText(Reader text, int pageLimit)
...
+         * </p>
+         * <p>
+         * 3. File system fonts
+         * </p>
+         * <p>
+         * 4. Transformer default font
+         * </p>
+         * <p>
+         * 5. PdfBox default font
+         * </p>
```
tpage-alfresco, line 435:
```suggestion
         * <ol>
         *     <li>Standard font map
         *     <li>Font Mappers
         *     <li>File system fonts
         *     <li>Transformer default font
         *     <li>PdfBox default font
         * </ol>

```

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
tiagosalvado10: Changed how exceptions are printed according to slf4j pattern.

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
```diff
@@ -130,6 +158,25 @@ public void transform(final String sourceMimetype, final String targetMimetype,
...
+        if (pdfFontSize != null && !pdfFontSize.isBlank())
+        {
+            try
+            {
+                fontSize = parseInt(pdfFontSize, PDF_FONT_SIZE);
+            }
+            catch (Exception e)
+            {
+                fontSize = DEFAULT_FONT_SIZE;
+                logger.error("Error parsing font size {}, going to set it as {} : {}", pdfFontSize, fontSize, e);
```
tpage-alfresco, line 177:
```suggestion
                logger.error("Error parsing font size {}, going to set it as {}", pdfFontSize, fontSize, e);

```

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
```diff
@@ -344,6 +403,199 @@ public PDDocument createPDFFromText(Reader text, int pageLimit)
...
+
+            if (mapping != null && mapping.getFont() != null && !mapping.isFallback())
+            {
+                try
+                {
+                    font = PDType0Font.load(doc, mapping.getFont().getOriginalData());
+                }
+                catch (Exception e)
+                {
+                    logger.error("Error loading font mapping {} : {}", fontName, e);
```
tpage-alfresco, line 512:
```suggestion
                    logger.error("Error loading font mapping {}", fontName, e);

```

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
```diff
@@ -344,6 +403,199 @@ public PDDocument createPDFFromText(Reader text, int pageLimit)
...
+                        if (null != fontIS)
+                        {
+                            PDDocument documentMock = new PDDocument();
+                            font = PDType0Font.load(documentMock, fontIS);
+                            break;
+                        }
+                    }
+                    catch (IOException ioe)
+                    {
+                        logger.error("Error loading font {} from filesystem: {}", fontName, ioe);
```
tpage-alfresco, line 551:
```suggestion
                        logger.error("Error loading font {} from filesystem", fontName, ioe);

```

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
```diff
@@ -344,6 +403,199 @@ public PDDocument createPDFFromText(Reader text, int pageLimit)
...
+                    {
+                        if (fontIS != null)
+                        {
+                            try
+                            {
+                                fontIS.close();
+                            }
+                            catch (Exception e)
+                            {
+                                logger.error("Error closing font inputstream: " + e.getMessage());
```
tpage-alfresco, line 563:
```suggestion
                                logger.error("Error closing font inputstream", e);

```

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
tiagosalvado10: Updated log statement.

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
tiagosalvado10: Updated log statement.

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/TextToPdfContentTransformer.java
tiagosalvado10: Updated log statement.

tpage-alfresco (approved):
LGTM - thanks Tiago.

PR #892, opened 2023-11-27
ACS-6151 - Create ATS liveness/readiness probes tests

goleksyalfresco:


engines/base/pom.xml
```diff
@@ -97,6 +97,19 @@
...
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
+            <version>1.19.1</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <version>1.19.1</version>
```
pzhyland, line 111: `<scope>test</scope>` seems to be missing here

goleksyalfresco: Added missing scope

engines/base/src/test/java/org/alfresco/transform/base/LivenessReadinessProbeTest.java
```diff
@@ -0,0 +1,107 @@
...
+        final ImagesForTests testData = getImageForTest();
+
+        try (final var env = createEnv(testData.image))
+        {
+            env.start();
+            var url = "http://localhost:" + env.getFirstMappedPort();
+
+            int max_transforms = 11;
+            for (int i = 0; i<max_transforms; i++) {
+                sendTransformRequest(url, testData.sourceMimetype, testData.targetMimetype, testData.filename);
```
pzhyland, line 37: Would be good to add an assertion that probe is successful before sending transformation request

engines/base/pom.xml
goleksyalfresco: Update: after adding the scope all tests has failed: `TestEngine with ID 'junit-jupiter' failed to discover tests`

engines/base/src/test/java/org/alfresco/transform/base/LivenessReadinessProbeTest.java
```diff
@@ -0,0 +1,107 @@
...
+    public void readinessShouldReturnAn429ErrorAfterReachingMaxTransforms() throws URISyntaxException
+    {
+        final ImagesForTests testData = getImageForTest();
+
+        try (final var env = createEnv(testData.image))
+        {
+            env.start();
+            var url = "http://localhost:" + env.getFirstMappedPort();
+
+            int max_transforms = 11;
```
pzhyland, line 35: Make `max_transforms` a class level constant. Set it to `10` and use it here and also in the `createEnv` method.

engines/base/src/test/resources/test.jpeg
pzhyland, line 1: Can this file be moved to the concrete TEngine using it?

engines/imagemagick/src/test/java/org/alfresco/transform/imagemagick/ImageMagickLivenessReadinessProbeTest.java
pzhyland, line 1: Can this file be moved to the concrete TEngine using it?

pzhyland (changes requested):
Some small changes are needed

engines/base/src/test/java/org/alfresco/transform/base/LivenessReadinessProbeTest.java
goleksyalfresco: Added an assertion before sending each request

engines/base/src/test/resources/test.jpeg
goleksyalfresco: Instead of moving this file I used existing quick.jpg and quick.pdf files for ImageMagick and PdfRenderer tests. Removed my test files.

engines/base/src/test/java/org/alfresco/transform/base/LivenessReadinessProbeTest.java
goleksyalfresco: Moved this value to a constant and used in both places

pzhyland (approved):

PR #956, opened 2024-05-14
ACS-6305 Fix Pipeline scan detecting 3rd party libraries

mikolajbrzezinski:
For some reason, after [this change](https://github.com/Alfresco/alfresco-transform-core/commit/7a2a7708672a65b595cb72747504b9e7deb399f3) Veracode started picking up bouncycastle and aspectjweaver as our own code instead of 3rd party library. Temporarily removing them from the package we send for scanning until this issue is resolved.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,6 +83,7 @@ jobs:
       - name: "Build"
         run: mvn -B -U install -DskipTests
       - name: "Create zip"
+        # Removing the aspectjweaver and bouncycastle jars from the scan, since Veracode detects them as 1st party code and fails the scan. TO BE REVERTED ONCE VERACODE FIXES THE ISSUE
         run:  |
           mkdir -p to-scan
           for file in engines/aio/target/alfresco-transform-core-aio-*.jar engines/base/target/alfresco-base-t-engine-*.jar model/target/alfresco-transform-model-*.jar
@@ -91,7 +92,6 @@ jobs:
               mv "$file" to-scan/
             fi
           done
-# Removing the aspectjweaver and bouncycastle jars from the scan, since Veracode detects them as 1st party code and fails the scan. TO BE REMOVED ONCE VERACODE FIXES THE ISSUE
           zip -d to-scan/alfresco-transform*.jar "BOOT-INF/lib/bcmail-jdk18on-*.jar" "BOOT-INF/lib/bcprov-jdk18on-*.jar" "BOOT-INF/lib/aspectjweaver*.jar"
           zip -r to-scan.zip to-scan
       - name: "Run SAST Scan"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,6 +92,7 @@ jobs:
             fi
           done
           zip -d to-scan/alfresco-transform*.jar "BOOT-INF/lib/bcmail-jdk18on-*.jar" "BOOT-INF/lib/bcprov-jdk18on-*.jar" "BOOT-INF/lib/asm*.jar"
+          zip -d to-scan/alfresco-base*.jar "BOOT-INF/lib/asm*.jar"
           zip -r to-scan.zip to-scan
       - name: "Run SAST Scan"
         uses: veracode/Veracode-pipeline-scan-action@v1.0.10
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,7 +98,6 @@ jobs:
           summary_output_file: results.json
           summary_display: true
           baseline_file: baseline.json
-          include: "*alfresco*"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -83,7 +83,17 @@ jobs:
...
-        run: zip -r to-scan.zip engines/aio/target/alfresco-transform-core-aio-*.jar engines/base/target/alfresco-base-t-engine-*.jar model/target/alfresco-transform-model-*.jar
+        run:  |
+          mkdir -p to-scan
+          for file in engines/aio/target/alfresco-transform-core-aio-*.jar engines/base/target/alfresco-base-t-engine-*.jar model/target/alfresco-transform-model-*.jar
+          do
+            if [[ $file != *javadoc.jar ]] && [[ $file != *sources.jar ]] && [[ $file != *tests.jar ]]; then
+              mv "$file" to-scan/
+            fi
+          done
+# Removing the aspectjweaver and bouncycastle jars from the scan, since Veracode detects them as 1st party code and fails the scan. TO BE REMOVED ONCE VERACODE FIXES THE ISSUE
```
dsibilio:
```suggestion
          # Removing the aspectjweaver and bouncycastle jars from the scan, since Veracode detects them as 1st party code and fails the scan. TO BE REMOVED ONCE VERACODE FIXES THE ISSUE
```

.github/workflows/ci.yml
```diff
@@ -98,6 +108,7 @@ jobs:
           summary_output_file: results.json
           summary_display: true
           baseline_file: baseline.json
+          include: "to-scan/alfresco-transform*"
```
dsibilio, line 111: Why did we have to add this? Won't this mess up the detection logic for other libraries that don't start with `alfresco-transform` such as e.g.: `alfresco-base-t-engine-*.jar` that is mentioned [within this very workflow file](https://github.com/Alfresco/alfresco-transform-core/pull/956/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR88).

.github/workflows/ci.yml
```diff
@@ -83,7 +83,17 @@ jobs:
       - name: "Build"
         run: mvn -B -U install -DskipTests
       - name: "Create zip"
-        run: zip -r to-scan.zip engines/aio/target/alfresco-transform-core-aio-*.jar engines/base/target/alfresco-base-t-engine-*.jar model/target/alfresco-transform-model-*.jar
+        # Removing the aspectjweaver and bouncycastle jars from the scan, since Veracode detects them as 1st party code and fails the scan. TO BE REVERTED ONCE VERACODE FIXES THE ISSUE
```
dsibilio, line 86: Let's move this comment down just above the actual shell line that does this: https://github.com/Alfresco/alfresco-transform-core/pull/956/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR95

.github/workflows/ci.yml
mikolajbrzezinski: The -include parameter only allows to choose from what already is recognized by SAST Scan as top module. In this case, this allows us to remove the jwarc.jar . 
If you think this is unnecessary, I can revert it.

Here is the effect of this line:

==========================
Found 2 Scannable modules.
==========================
JS files within jwarc-0.29.0.jar
to-scan/alfresco-transform-core-aio-5.1.2-A2-SNAPSHOT.jar

===================
Analyzed 1 modules.
===================
to-scan/alfresco-transform-core-aio-5.1.2-A2-SNAPSHOT.jar

.github/workflows/ci.yml
mikolajbrzezinski: For some reason, adding a comment inside of run command was causing a syntax error, which I fixed by moving the comment to where it is now
https://github.com/Alfresco/alfresco-transform-core/actions/runs/9092835071/workflow
https://github.com/Alfresco/alfresco-transform-core/actions/runs/9092752774/workflow

.github/workflows/ci.yml
dsibilio: We can achieve the very same result by having a more lax include filter such as `to-scan/alfresco-*` then. Let's try that instead.

.github/workflows/ci.yml
dsibilio: I reckon that's because it was formatted badly, the comment has to start at the same indentation level as the rest of the shell commands, not at the start of the line. This will mess up the YAML interpreter.

You can most definitely add comments in the Shell script that's run via `run: |` without any issues.

dsibilio (approved):

damianujma (approved):

PR #961, opened 2024-05-28
[ACS-5648] Add recovery mode for TransformRegistry

kmagdziarz:


engines/base/src/main/java/org/alfresco/transform/base/registry/TransformRegistry.java
```diff
@@ -151,44 +142,77 @@ public void handleContextRefreshedEvent(final ContextRefreshedEvent event)
      * Load the registry on application startup. This allows Components in projects that extend the t-engine base
      * to use @PostConstruct to add to {@code transformConfigSources}, before the registry is loaded.
      */
```
kmagdziarz, line 144: Since I added IllegalStateException handling in retrieveConfig() there is no longer need for @retryable, @async annotation was needed to enable retry in asynchronous way.

pzhyland: Can we still keep it `@async`?

engines/base/src/main/java/org/alfresco/transform/base/registry/TransformRegistry.java
```diff
@@ -151,44 +142,77 @@ public void handleContextRefreshedEvent(final ContextRefreshedEvent event)
...
+        retrieveConfig();
     }
 
     /**
-     * Takes the schedule from a spring-boot property
+     * Recovery mode method, executes only when there was failure during initial config retrieval process.
      */
-    @Scheduled(cron = "${transform.engine.config.cron}")
-    public void retrieveEngineConfigs()
+    @Scheduled(initialDelay = 1000, fixedDelayString = "#{${transform.engine.config.retry.timeout} * 1000}")
```
pzhyland, line 164: 1s might be not enough to give time for the initmethod. I suggest setting it to `#{${transform.engine.config.retry.timeout} * 1000}`

engines/base/src/main/java/org/alfresco/transform/base/registry/TransformRegistry.java
kmagdziarz: Unfortunately no - it was added for @Retryable, not the other way around.

pzhyland: Are you sure? I am quite confident that the `@Async` does not require the `@Retryable`.

kmagdziarz: You are right. I meant that @Retryable needed @Async. But as we discovered we can still keep @Async.

pzhyland (approved):

PR #1078, opened 2025-04-24
AAE-34154 Bump Apache POI to 5.4.1

alep85:
https://hyland.atlassian.net/browse/AAE-34154


dsibilio (changes requested):
Looks like tests are failing and may be related to this change.

For reference, when I had to bump this in the past it worked best when also bumping other dependencies that are related such as Tika and PDFBox: https://github.com/Alfresco/alfresco-transform-core/commit/b4dd80de05c363073f6b28f057c5b47b361eaf02.

PR #1079, opened 2025-05-22
Fix/mnt 25089 html transformations with ootb aio create extra whitespace

kushal-banik-hyland:
This pull request introduces a new feature for HTML-to-text transformation by adding a "collapseHtml" option (`MISC_HTML_COLLAPSE`). It also includes updates to documentation, configuration files, and tests, along with minor refactoring and copyright updates.

### New Feature: HTML Collapse Option
* Added `MISC_HTML_COLLAPSE` property to enable or disable collapsing of HTML content during transformation. This property is configurable via `application-default.yaml` and defaults to `true`.
* Updated `HtmlParserContentTransformer` to respect the `MISC_HTML_COLLAPSE` setting, with fallback to a default value if not explicitly provided. 
* Modified `misc_engine_config.json` to include `htmlOptions` with the `collapseHtml` parameter.

### Documentation Updates
* Updated `docs/external-engine-configuration.md` to document the new `MISC_HTML_COLLAPSE` property and its default value `true`.
### Test Enhancements
* Adjusted test file references and parameters to align with the new feature. 

These changes collectively enhance the flexibility and maintainability of the transformation engine while ensuring compatibility with existing functionality.


engines/misc/src/main/java/org/alfresco/transform/misc/MiscTransformEngine.java
```diff
@@ -74,6 +75,6 @@ public TransformConfig getTransformConfig()
     public ProbeTransform getProbeTransform()
     {
         return new ProbeTransform("probe.html", MIMETYPE_HTML, MIMETYPE_TEXT_PLAIN, transformOptions,
-                119, 30, 150, 1024, 60 * 2 + 1, 60 * 2);
+                107, 30, 150, 1024, 60 * 2 + 1, 60 * 2);
```
damianujma, line 78: Why this change is needed?

kushal-banik-hyland: Actually in this  probe it checked some size for the transformed file for Misc Engine. The total size was 89 previously (with spaces). Which is from the range `119 - 30 = 89` to `119 + 30 = 149`  but now as there is no space in the file, size is `87`. So I have modified to have in the boundary in `77 to 137`.

tiagosalvado10 (changes requested):
Since the `html` transformer (from `misc` engine) is currently working with the `collapse` option set to `false` I would suggest to make this option configurable (just in case a customer wants to keep relying on the previous behavior instead of the new one).

With this change, the customer would be able to choose whether to collapse or not during the HTML transformation:

1. In `engines/misc/src/main/resources/misc_engine_config.json` file:
1.1. Add `htmlOptions` to `transformOptions` section with value `collapse`
1.2. Change the `html` transformer to use the above `htmlOptions`
2. In `org.alfresco.transform.misc.transformers.HtmlParserContentTransformer` java class:
2.1. Obtain the `collapse` value from `transformOptions` map (it can be `true` as default)
3. Add unit tests to cover both scenarios

kushal-banik-hyland:
Currently added the feature to have dynamic transform option for HTML to TXT in `MISC Engine`, which can be set through `misc_engine_config.json` based on that anyone can select which behavior they need.

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/HtmlParserContentTransformer.java
```diff
@@ -1,203 +1,229 @@
...
+ */
+@Component
+public class HtmlParserContentTransformer implements CustomTransformerFileAdaptor
+{
+    private static final Logger logger = LoggerFactory.getLogger(
+            HtmlParserContentTransformer.class);
+
+    private static final String HTML_COLLAPSE = "collapseHtml";
+    private static final String HTML_OPTIONS = "htmlOptions";
+    private boolean COLLAPSE_HTML = true;
```
tiagosalvado10, line 86: This constant is not required here.

engines/misc/src/main/java/org/alfresco/transform/misc/transformers/HtmlParserContentTransformer.java
```diff
@@ -1,203 +1,229 @@
...
+    private MiscTransformEngine transformEngine;
+
+    @Override
+    public String getTransformerName()
+    {
+        return "html";
+    }
+
+    @PostConstruct
+    public void init() {
```
tiagosalvado10, line 98: This initialization can be removed since the option value will be obtained at request level, not at bean initialization.

engines/misc/src/test/java/org/alfresco/transform/misc/transformers/HtmlParserContentTransformerTest.java
```diff
@@ -1,162 +1,242 @@
...
+            }
+        }
+    }
+
+    /**
+     * Tests the transformer with different collapsing methods. If the collapsing is set to false, it should not collapse the new lines between paragraphs. If the collapsing is set to true, it should collapse the new lines.
+     */
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    public void testTransformerWithDifferentCollapsingMethods(boolean shouldCollapse)
```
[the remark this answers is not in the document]
kushal-banik-hyland: Done

model/src/main/java/org/alfresco/transform/common/RequestParamMap.java
```diff
@@ -1,93 +1,96 @@
...
+    String ENDPOINT_TRANSFORM_CONFIG_LATEST = ENDPOINT_TRANSFORM_CONFIG + "?" + CONFIG_VERSION + "=" + CONFIG_VERSION_LATEST;
+    String ENDPOINT_VERSION = "/version";
+    String ENDPOINT_READY = "/ready";
+    String ENDPOINT_LIVE = "/live";
+    String ENDPOINT_ERROR = "/error";
+    String ENDPOINT_LOG = "/log";
+    String ENDPOINT_ROOT = "/";
+
+    // Html parameter names for the transform config
+    String HTML_COLLAPSE = "collapseHtml";
```
tiagosalvado10, line 95: Just a minor change for consistency purposes: could this variable be placed below the `PDF_FONT_SIZE` one?

tiagosalvado10 (changes requested):

PR #1083, opened 2025-06-16
[AAE-34971] Bump tomcat embed dependencies version to 11.0.8

aonych:
https://hyland.atlassian.net/browse/AAE-34971


mikolajbrzezinski (approved):

aonych:
On Friday, we updated Spring Boot to 3.4.5 version, so I think there is no longer a problem with it. I'll update the spring boot version instead of the tomcat only. :) 
Thank you!

PR #1148, opened 2025-10-14
ACS-10404 Test e2e AIO runtime compatibility with Java 21 and 25

b-ansari:


_ci/e2e_aioTest.sh
gerardolenski: This file is just the exact copy of existing `test.sh` Please rather parametrize correctly the existing one. We don't want to have one more script to maintain if it is not necessary.

gerardolenski (changes requested):
I've added few comments to improve the solution. Also one additional regarding changes in Dockerfiles done on base branch:

- Some Dockerfiles contain additional java configuration for architecture different that `x86_64` (e.g. in aio you can find `update-alternatives --set java java-17-openjdk.aarch64;` This should be also correctly parametrized when newer Java version was set.

.github/workflows/ci.yml
```diff
@@ -238,10 +238,52 @@ jobs:
         timeout-minutes: ${{ fromJSON(env.GITHUB_ACTIONS_DEPLOY_TIMEOUT) }}
         run: bash _ci/test.sh ${{ matrix.testProfile }}
 
+  e2e_javaruntime_tests:
+    name: "E2E AIO Tests : ${{ matrix.testName }} - ${{ matrix.java }}"
```
gerardolenski: 1. Pls rename it to `E2E AIO Smoke Tests (Java: x)`
2. the `${{ matrix.testName }}` is not used anymore

_ci/java_base_image_map.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+if [ -n "$JAVA_RUNTIME_VERSION" ]; then
+  declare -A JAVA_IMAGE_MAP=(
+    [21]="alfresco/alfresco-base-java:jre21-rockylinux9@sha256:27297bffe7d45152194f08a8f38917a7e3d43b9a34dcfcdc29b83a80c4c33a10"
```
gerardolenski: New images were released 4 days ago, pls udate

PR #1234, opened 2026-04-14
ACS-10652 fixing Test Failures in ACS

b-ansari:


model/src/test/java/org/alfresco/transform/registry/OverrideTransformConfigTests.java
```diff
@@ -83,6 +86,7 @@ public class OverrideTransformConfigTests
             .withSourceMediaType("mimetype/x")
             .withTargetMediaType("mimetype/y")
             .withMaxSourceSizeBytes(200L)
+            .withPriority(23)
```
gerardolenski, line 91: I have a feeling you changed existing ATS configuration behavior.
In this test looks like someone wanted to verify if:
- ⁠supported_A2B__40 - override config with priority but without source bytes field defined
- ⁠supported_X2Y_200 - override config with source size byte but without priority field defined 

Now you changed it, so is that mean we do not verify anymore such situation?

The tests in community-repo failed when this configuration was applied:
```
{
  "overrideSupported": [
    {
    "transformerName": "appleIWorks",
    "sourceMediaType": "application/vnd.apple.pages",
    "targetMediaType": "image/jpeg",
    "maxSourceSizeBytes": 786432
    }
  ],
  "addSupported": [
    {
      "transformerName": "imagemagick",
      "sourceMediaType": "image/jpeg",
      "targetMediaType": "rubbish"
    }
  ]
} 
```

b-ansari: The test field definitions are updated to match the new semantics:

supported_A2B__40 gained [withMaxSourceSizeBytes(-1L)] — because now the default-resolved size (-1, meaning unlimited) is retained instead of reset.

supported_X2Y_200 gained [.withPriority(23)]— because now the existing priority 23 is retained instead of reset.

So yes — the test was updated to reflect the new behavior, not to preserve the original contract. The scenario "unspecified fields in an override should be cleared" is no longer tested.

gerardolenski: All right. Just modify as we discussed to avoid confusion in these tests.

gerardolenski (approved):
LGTM

model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
```diff
@@ -218,23 +223,83 @@ private void addSupported(Set<AddSupported> addSupportedSet, String readFrom, Ab
...
+                SupportedSourceAndTarget existingSupported = existingSupportedOpt.get();
+                supportedList.remove(existingSupported);
+                if (override.getMaxSourceSizeBytes() != null)
+                {
+                    existingSupported.setMaxSourceSizeBytes(override.getMaxSourceSizeBytes());
+                }
+                if (override.getPriority() != null)
+                {
+                    existingSupported.setPriority(override.getPriority());
+                }
```
b-ansari, line 292: New change added as part of ACS test failures

sobanerjee-hyland (approved):

PR #1243, opened 2026-04-24
ACS-10652 Changing stepTransformer overrride logic to fix Test Failures inACS

b-ansari:


model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
```diff
@@ -258,48 +258,98 @@ private void applyDeferredOverrides(AbstractTransformRegistry registry)
...
+                        supported.getTargetMediaType().equals(override.getTargetMediaType()))
+                .findFirst();
+
+        if (existingSupportedOpt.isPresent())
+        {
+            applyOverrideValuesToEntry(override, existingSupportedOpt.get(), supportedList);
+        }
+        else
+        {
+            leftoverBySource.computeIfAbsent(readFrom, k -> new HashSet<>()).add(override);
```
gerardolenski: Why this method modifies the state of `leftoverBySource` map? It is very confusing. Please change this design to correctly return processing result that can be handled by parent method.

model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
```diff
@@ -258,48 +258,98 @@ private void applyDeferredOverrides(AbstractTransformRegistry registry)
...
+    {
+        Set<SupportedSourceAndTarget> supportedList = transformer.getSupportedSourceAndTargetList();
+        Optional<SupportedSourceAndTarget> existingSupportedOpt = supportedList.stream()
+                .filter(supported -> supported.getSourceMediaType().equals(override.getSourceMediaType()) &&
+                        supported.getTargetMediaType().equals(override.getTargetMediaType()))
+                .findFirst();
+
+        if (existingSupportedOpt.isPresent())
+        {
+            applyOverrideValuesToEntry(override, existingSupportedOpt.get(), supportedList);
```
gerardolenski: I quite don't understand why you implemented it this way. You found `existingSupported` in `supportedList` and then inside `applyOverrideValuesToEntry` you are removing it from the set, then adding it again. Why just not modify the `MaxSourceSizeBytes` and `Priority` instead?

model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
```diff
@@ -258,48 +258,98 @@ private void applyDeferredOverrides(AbstractTransformRegistry registry)
...
             {
-                leftoverBySource.computeIfAbsent(readFrom, k -> new HashSet<>()).add(override);
+                continue;
             }
+
+            Set<SupportedSourceAndTarget> supportedList = pipeline.getSupportedSourceAndTargetList();
+            supportedList.stream()
+                    .filter(supported -> override.getSourceMediaType().equals(supported.getSourceMediaType()))
+                    .collect(Collectors.toList())
+                    .forEach(existingSupported -> applyOverrideValuesToEntry(override, existingSupported, supportedList));
```
gerardolenski: What is happening here? You are iterating `supportedList` (which is `Set` btw) and inside `forEach` you are removing and adding entries of this collection.

gerardolenski (changes requested):
See comments.

model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
b-ansari: Hi,

[hashCode()] includes [maxSize] and [priority].
The objects are stored in a [HashSet<SupportedSourceAndTarget>]. If we mutate a field that [hashCode()] depends on while the object is still in the set, the object ends up in the wrong bucket — it becomes effectively invisible to lookups and you get a corrupted set.

**remove** — take it out of the set (uses the current correct hash to find it)
**modify [maxSourceSizeBytes] / [priority]**— fields that affect hash change
**[add]** — put it back with the new hash in the right bucket
If we just did:
```
 existingSupported.setMaxSourceSizeBytes(...);
 existingSupported.setPriority(...);
 // object is now in wrong bucket — set is corrupted
```

gerardolenski: So why not calculate new `Set` for `Transformer` instead?

model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
b-ansari: yes what i can do here is separate the concerns: make applyOverrideValuesToEntry only modify the object (not touch the set), and let callers own the remove/re-add explicitly.

model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
b-ansari: updated

model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
```diff
@@ -258,50 +257,102 @@ private void applyDeferredOverrides(AbstractTransformRegistry registry)
...
+            // --- Apply to the directly-named transformer (exact source + target match) ---
+            boolean applied = applyOverrideToDirectTransformer(override, directMatches.get(0));
+            if (!applied)
             {
                 leftoverBySource.computeIfAbsent(readFrom, k -> new HashSet<>()).add(override);
             }
+
+            // --- Propagate to any pipeline whose first step is the overridden transformer ---
+            propagateOverrideToPipelineParents(override);
         }
```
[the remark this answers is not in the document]
b-ansari: Fixed

model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
```diff
@@ -258,50 +257,102 @@ private void applyDeferredOverrides(AbstractTransformRegistry registry)
...
+            {
+                continue;
+            }
+
+            pipeline.setSupportedSourceAndTargetList(
+                    pipeline.getSupportedSourceAndTargetList().stream()
+                            .map(entry -> override.getSourceMediaType().equals(entry.getSourceMediaType())
+                                    ? withOverrideApplied(override, entry)
+                                    : entry)
+                            .collect(toSet()));
```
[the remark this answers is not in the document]
b-ansari: updated

model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
```diff
@@ -258,50 +257,102 @@ private void applyDeferredOverrides(AbstractTransformRegistry registry)
...
+            if (!override.getTransformerName().equals(steps.get(0).getTransformerName()))
+            {
+                continue;
+            }
+
+            pipeline.setSupportedSourceAndTargetList(
+                    pipeline.getSupportedSourceAndTargetList().stream()
+                            .map(entry -> override.getSourceMediaType().equals(entry.getSourceMediaType())
+                                    ? withOverrideApplied(override, entry)
+                                    : entry)
```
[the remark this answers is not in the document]
b-ansari: updated

model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
```diff
@@ -258,50 +258,117 @@ private void applyDeferredOverrides(AbstractTransformRegistry registry)
...
+            supportedList.addAll(entriesToOverride.stream()
+                    .map(entry -> withOverrideApplied(override, entry))
+                    .collect(toSet()));
+        }
+    }
+
+    /**
+     * Returns a new {@link SupportedSourceAndTarget} copied from {@code existing} with non-null override fields applied. Null override fields are left unchanged to preserve values already set by applyDefaults().
+     */
+    private SupportedSourceAndTarget withOverrideApplied(OverrideSupported override, SupportedSourceAndTarget existing)
```
gerardolenski: - The name of method should be a verb, right? Maybe `applyOverride`, `mergeOverride` or sth simillar?
- `existing` name is a bit strange for this method signature
- the metthod is taking given `upportedSourceAndTarget `and adds to is `OverrideSupported` - so logically first parameter in signature should be `SupportedSourceAndTarget`

model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
```diff
@@ -258,50 +258,117 @@ private void applyDeferredOverrides(AbstractTransformRegistry registry)
...
+    /**
+     * Propagates the override to every pipeline transformer whose first step is the overridden transformer. Matches on both source media type and the step's target media type (the intermediate type) to avoid incorrectly updating pipelines that share the same first-step transformer name but use a different intermediate type. The synthesised pipeline entry target is the final pipeline output, not the intermediate, so only the source and the step's own target are used for matching.
+     */
+    private void propagateOverrideToPipelineParents(OverrideSupported override)
+    {
+        for (Origin<Transformer> origin : combinedTransformers)
+        {
+            Transformer pipeline = origin.get();
+            List<TransformStep> steps = pipeline.getTransformerPipeline();
+            if (steps == null || steps.isEmpty())
```
gerardolenski: Why not using `CollectionUtils::isEmpty` ?

model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
```diff
@@ -258,50 +258,117 @@ private void applyDeferredOverrides(AbstractTransformRegistry registry)
...
+            {
+                continue;
```
gerardolenski: Do we really needs this ugly `continue` statements? I'm sure you could refactor it.

model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
```diff
@@ -258,50 +258,117 @@ private void applyDeferredOverrides(AbstractTransformRegistry registry)
...
         leftoverBySource.forEach((readFrom, leftOvers) -> logWarn(leftOvers, readFrom, registry, "overrideSupported"));
         deferredOverrides.clear();
     }
 
+    /**
+     * Applies the override to the entry on the directly-named transformer that matches the override's source and target media types exactly.
+     *
+     * @return {@code true} if a matching entry was found and updated; {@code false} if no match was found (caller should treat the override as unresolved)
+     */
+    private boolean applyOverrideToDirectTransformer(OverrideSupported override, Transformer transformer)
```
gerardolenski: Logically, you are taking `Transformer` and applying to it `OverrideSupported`. The parameters order could follow this.

model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
```diff
@@ -258,50 +258,117 @@ private void applyDeferredOverrides(AbstractTransformRegistry registry)
...
+    {
+        Set<SupportedSourceAndTarget> existing = transformer.getSupportedSourceAndTargetList();
```
gerardolenski: Thisnk about better name for `existing` variable. This name is vague. Maybe `currentEntries`, `originalEntries` or even `existingEntries`?

model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
```diff
@@ -258,50 +258,117 @@ private void applyDeferredOverrides(AbstractTransformRegistry registry)
...
+
+        List<SupportedSourceAndTarget> entriesToOverride = existing.stream()
+                .filter(entry -> Objects.equals(override.getSourceMediaType(), entry.getSourceMediaType()) &&
+                        Objects.equals(override.getTargetMediaType(), entry.getTargetMediaType()))
+                .collect(Collectors.toList());
+
+        replaceEntriesWithOverrides(existing, entriesToOverride, override);
```
gerardolenski: So if `replaceEntriesWithOverrides` has logic to conditionally override, shouldn't it return boolean in their signature?

b-ansari: Yes, added Thankyou

model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
b-ansari: Sure, updated

model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
b-ansari: NA

model/src/main/java/org/alfresco/transform/registry/CombinedTransformConfig.java
b-ansari: Yes, Correct.
Updated

gerardolenski (approved):
Ok. Let it be. However I have got feeling that the entire flow of combined transform config is overcomplicated. It might be good candidate for the story for the future.

arindam-roy-hld (approved):

PR #1319, opened 2026-07-27
[ACS-12260]: Automated alfresco pdf rendered updates

Vivekkr0311:
# Jira: https://hyland.atlassian.net/browse/ACS-12260

## Purpose

`alfresco-pdf-renderer` publishes a new release tag every **Sunday (00:00 UTC)**. Until now, bumping the pinned `ARG PDF_RENDERER_VERSION` in this repo's Dockerfiles was a manual, easy-to-forget chore. This PR adds a **scheduled CI workflow that detects the new release and opens (or updates) a single pull request automatically**, keeping the `pdfrenderer` and `aio` images current with zero manual effort while still requiring human review before anything reaches `master`.

## What's added

| File | Role |
|---|---|
| `.github/workflows/update-pdf-renderer.yml` | Orchestrates the whole flow |
| `_ci/update-pdf-renderer/check-version.sh` | Detects whether a newer release exists |
| `_ci/update-pdf-renderer/update-dockerfiles.sh` | Rewrites the version in both Dockerfiles |

Logic is split into shell scripts (not inline YAML) so each step is testable and readable in isolation.

## How it works

**Triggers**
- **Scheduled**: `cron: '0 2 * * 1'` — every **Monday 02:00 UTC**. Deliberately timed *after* the pdf-renderer Sunday release and ~3h *before* this repo's own Monday 05:00 UTC CI.
- **Manual**: `workflow_dispatch`.

**Step-by-step**
1. **Mint a scoped token** — `actions/create-github-app-token` issues a short-lived GitHub App token scoped to just `alfresco-transform-core` + `alfresco-pdf-renderer`, with only `contents: write` and `pull-requests: write`. Workflow-level default is `contents: read` (least privilege).
2. **Check for a new version** (`check-version.sh`) — lists tags from `Alfresco/alfresco-pdf-renderer`, filters to `X.Y.Z-N` format, `sort -V | tail -1` to find the latest. Reads the current `ARG PDF_RENDERER_VERSION` from **both** Dockerfiles and **fails fast if they disagree**. Emits `latest` and `changed` (true/false) outputs.
3. **Update Dockerfiles** (`update-dockerfiles.sh`, only if `changed==true`) — `sed`-replaces the pinned version in `engines/pdfrenderer/Dockerfile` and `engines/aio/Dockerfile`, then prints the diff.
4. **Open/update the PR** — `peter-evans/create-pull-request` (SHA-pinned to v8.1.1) commits to a **fixed branch** `auto/update-pdf-renderer`, scoped to just the two Dockerfiles, with a descriptive title/body. Because the branch name is fixed, a later run **updates the existing open PR in place** instead of spawning a new one — so there are never stale/duplicate auto-PRs to clean up. Normal CI runs on the PR and validates the new version before merge.

**Reviewer assignment**
- **Manual run** → the user who triggered it (`github.actor`) is set as reviewer.
- **Scheduled run** → the `alfresco-dependency-maintainers` team is requested for review.

Auto-PRs are **never auto-merged** — final merge to `master` is a manual decision by the maintainers team.


.github/workflows/update-pdf-renderer.yml
```diff
@@ -0,0 +1,83 @@
...
+            Releases: https://github.com/Alfresco/alfresco-pdf-renderer/releases
+
+            CI on this PR validates the new version before it is merged to `master`.
+          add-paths: |
+            engines/pdfrenderer/Dockerfile
+            engines/aio/Dockerfile
+          reviewers: ${{ github.event_name == 'workflow_dispatch' && github.actor || '' }}
+          team-reviewers: ${{ github.event_name == 'schedule' && 'alfresco-dependency-maintainers' || '' }}
+
+      - name: Close stale pull requests
```
gionn: don't use a dynamic branch in the previous step and you are not going to have stale pull requests

Vivekkr0311: As discussed with @kmagdziarz , we should also close the PR which were not merged by mistake.

@kmagdziarz Please correct me if I miss understood anything with this thing.

Vivekkr0311: Thanks Giovanni, You are correct, I should remove the dynamic branches, as it could populate lot of branches.
But for automatic closing PR, I need guidance, how we can approach this.

gionn: just configure a fixed string for the branch in the peter-evans/create-pull-request action, it will automatically update the existing pr if it's already there and adding the newly modified files

check the docs:
https://github.com/peter-evans/create-pull-request#-create-pull-request

gionn (approved):

PR #329, opened 2021-02-02
ATS-851: Improve build reliability

TheOutsSeeker:
- Add Maven Central to top of repo search list
- Disable Maven connection pool

PR #1330, opened 2026-08-04
Skip snapshot deploy on non-master CI builds

gionn:
## Summary

The `build_and_test` job ("Core & Base Snapshot deployment") runs `mvn clean deploy -Pbase` unconditionally, which is why Dependabot PRs fail CI with a 403 from `artifacts.alfresco.com` — Dependabot resolves read-only Nexus credentials for the PR context, since dependabot.yml only needs them for version resolution. This is a regression from #1280, which replaced the job's original `github.ref_name == 'master' && github.event_name != 'pull_request'` guard with the shared `test_run_check` gate, so the deploy step now runs on every PR, not just master pushes.

Worse, human-authored PRs don't fail: they carry write-capable Nexus secrets, so the deploy succeeds and publishes over the shared `5.4.5-A.1-SNAPSHOT` line from a PR branch (confirmed in a recent successful PR run, which uploaded `alfresco-transform-core`, `alfresco-transform-model` and `alfresco-base-t-engine` snapshots). Dependabot's read-only token just surfaced a pre-existing problem.

This change derives the Maven phase (`deploy` vs `verify`) from the event/branch context in the "Run tests" step, matching the existing pattern already used in `_ci/test.sh` for the other test jobs (which is why `all_tests_matrix` and the E2E jobs don't fail on Dependabot PRs). Snapshot deploys now only happen on `master` pushes; PRs (including Dependabot's) still build and run the `-Pbase` tests via `verify`, with no Nexus write. Job and step names are also updated since the job no longer always deploys.

Also drops `-Dmaven.wagon.http.pool=false` from every CI/release `mvn` invocation (`ci.yml`, `_ci/build.sh`, `_ci/test.sh`, `_ci/release.sh`, `_ci/buildAndTestAll.sh`). It was introduced in 2021 (ATS-851, #329) as a workaround for wagon-http connection-reuse failures against Nexus and then copy-pasted into every later entrypoint without being revisited. Maven/wagon has been considerably more stable since, especially over the last year, and current CI logs show no trace of the failure it was working around, so it's being retired rather than propagated further.

## Test plan

- [x] YAML validated (`python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"`)
- [x] Shell scripts pass `bash -n` after removing the flag
- [ ] Draft PR CI run confirms the "Core & Base" job resolves to `mvn ... clean verify ... -Pbase` with no `Uploading to alfresco-public-snapshots` lines
- [ ] After merge, a master push still logs `Uploaded to alfresco-public-snapshots: .../alfresco-transform-core-<version>.pom`
- [ ] Re-run checks on an open Dependabot PR to confirm it now goes green


kmagdziarz (dismissed):

kmagdziarz:
https://github.com/Alfresco/alfresco-transform-core/pull/1321 this may be in conflict with this. @gionn

gionn:
will wait for it to merge

dsibilio (dismissed):

damianujma (dismissed):

gionn:
@copilot resolve the merge conflicts in this pull request

damianujma (approved):

